PR #15, opened 2015-06-16
Initial commit to add support for windows universal platform

rakatyal:


rakatyal:
You are right. Corrected.

Issue #42, opened 2015-07-15
iOS only receives first notification in foreground


Deminetix:
Sending a push notification only calls on('notification') once, while the app is in the foreground.

To reproduce:

- add a console.log to the `on('notification')` callback
- deploy to device and leave the app in the foreground
- send it a push notification
- XCode logs native `Push Plugin key: alert` etc
- `on('notification')` callback is fired and outputs to log
- send another push notification
- xcode still logs native `Push Plugin key: alert` etc
- on('notification') doesn't get fired

Deminetix:
Figured it out. The reason this happens is because the new `setApplicationBadgeNumber` functionality overwrites the callbackId used for the `.on()` methods.

I'll make a pull request for this and #41 soon but for those who need it here's what i did.

1. I changed the javascript prototype method for `setApplicationBadgeNumber` to use `exec()` instead of `cordova.exec()` as it's already required in the file.

```
exec(successCallback, errorCallback, "PushNotification", "setApplicationIconBadgeNumber", [{badge: badge}]);
```

2. I changed `-(void)setApplicationIconBadgeNumber...` to use his own callbackId instead of overwriting the `.on()` pipe callbackId

OLD:

```
- (void)setApplicationIconBadgeNumber:(CDVInvokedUrlCommand *)command {

    self.callbackId = command.callbackId;

    NSMutableDictionary* options = [command.arguments objectAtIndex:0];
    int badge = [[options objectForKey:@"badge"] intValue] ?: 0;

    [[UIApplication sharedApplication] setApplicationIconBadgeNumber:badge];

    [self successWithMessage:[NSString stringWithFormat:@"app badge count set to %d", badge]];
}
```

NEW:

```
- (void)setApplicationIconBadgeNumber:(CDVInvokedUrlCommand *)command 
{
    NSMutableDictionary* options = [command.arguments objectAtIndex:0];
    int badge = [[options objectForKey:@"badge"] intValue] ?: 0;

    [[UIApplication sharedApplication] setApplicationIconBadgeNumber:badge];

    NSString* message = [NSString stringWithFormat:@"app badge count set to %d", badge];
    CDVPluginResult *commandResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK messageAsString:message];
    [self.commandDelegate sendPluginResult:commandResult callbackId:command.callbackId];
}
```

PR #43, opened 2015-07-15
fix issue #42

Deminetix:
This fixes issue #42 where `setApplicationIconBadgeNumber` was overwriting the callbackId pipe for the `.on()` method


macdonst:
@Deminetix Thanks so much, I'll test this shortly.

PR #44, opened 2015-07-15
fix issue #41

### LINKED ISSUE #41: iOS doesn't add foreground key
Android sets `additionalData.foreground: true` on notification data when the app is open. This is helpful because background notifications - in most apps - when tapped will go to a route... but if the app is open  we will likely alert the user in a different manner instead of route them away from where they are.

Is this not 100% necessary for iOS too? Maybe i'm missing another way to achieve the same thing?

Deminetix:
This fixed #41 and adds a foreground boolean the same as on Android

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Callback Parameter | Description
 `data.sound` | `String` The name of the sound file to be played upon receipt of the notification.
 `data.image (windows only)` | `String` The path of the image file to be displayed in the notification.
 `data.additionalData` | `JSON Object` An optional collection of data sent by the 3rd party push service that does not fit in the above properties.
+`data.additionalData.foreground` | `Boolean` Whether the notification was received while the app was in the foreground
 
 #### Example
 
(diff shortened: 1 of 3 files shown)


macdonst:
@Deminetix can you modify your fix so that is sets foreground to true when isInline is true and sets foreground to false when isInline is false. Then I'll test your PR and merge.

Deminetix:
@macdonst i'm not sure i follow. What was the original intention of the isInline variable?

macdonst:
I don't even think we need to be this fancy. The 'notificationReceived' method only gets called when the app is the the foreground.

macdonst:
@Deminetix okay, I merged this in but I will do the quick addition to set foreground to false myself. 

I really appreciate you stepping up to help out with this plugin. Finally a minor point, check out Chris' slides on commit messages. Something like this PR could be rebased so it is all one commit instead of 3.

https://github.com/CSilivestru/git-different

Issue #20, opened 2015-06-19
Android: Notification icon problem


sebrojas14:
In lollipop, My notification look like this:

In the notification bar, it's a white square (with the old plugin, the same app icon was used here):

![screenshot_2015-06-19-13-36-53](https://cloud.githubusercontent.com/assets/1771574/8260337/99415342-1688-11e5-95b1-6b2536210cf6.png)

Notification bar expanded, a circle with the app icon inside (there's no way make somethinh better, like the other apps?):

![screenshot_2015-06-19-13-37-00](https://cloud.githubusercontent.com/assets/1771574/8260352/a901b31c-1688-11e5-9e6d-e10d069649ad.png)

macdonst:
@sebrojas14 upon having the chance to look into this bug I believe it not to be a bug. In Android Lollipop there were changes to the notifications so all icons show up as white. See the following url: https://developer.android.com/design/patterns/notifications.html

I think the best way forward for you is to add a transparency to your icon so that it your icon shows up as a silhouette of a person. Potentially we need to add another icon into cordova/phonegap.

sebrojas14:
Yes, you're right. But I need the app icon to have a white background. So, you would be nice a a way to specify a different icon for notifications, or even better, the option of use icons with badge like skype and inbox in the image

macdonst:
@sebrojas14 so the way that it works on Android is you can specify a small icon and large icon in the notification. As you can see in this screen shot:

![Screenshot](https://dl.dropboxusercontent.com/u/887989/2015-06-22%2017.27.08.png)

You can see the PhoneGap Enterprise is the large icon and the cordova build bot is the small icon. I can add in the large icon support but we don't have a good way of specifying it right now. I'm changing this to an enhancement for now.

sebrojas14:
So for now, is not possible to specify a different icon for the notifications?

macdonst:
@sebrojas14 no, not at the moment but my plan is to add a option in the android setup to provide an icon to use instead of the main app icon. If not icon is provide it will use the main app icon as a fallback.

sebrojas14:
I'll wait for it! thank you

ivanmayes:
+1 this would be extremely helpful, white square are no fun!

michael-may:
+1

Tested the TargetSDK workaround and it brings the full icon back for now at least.

In your config.xml:
<platform name="android">
     ...
     <preference name="android-targetSdkVersion" value="19"/>
</platform>

[...]

PR #45, opened 2015-07-16
Android icon support

Deminetix:
This PR fixes the android lollipop issue displaying blank icons in issue #20 in a similar manner to what @macdonst mentioned

It adds support for both small-icon and large-icon, and gracefully falls back to how the plugin works prior to this PR if you don't set the android options

Small Icons:
- add the `icon` key to your android options which should be a string name of the resource to use (in res/drawables etc)
- if no android option is set, it falls back to what it currently does without this PR
- you can also override the android option key from the server by sending an `icon` key to GCM

Large Icons:
- add the `largeIcon` key to your android options which should be a string name of the resource to use (in res/drawables etc)
- if no android option is set, it does nothing
- you can also override the android option key from the server by sending a `largeIcon` key to GCM

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Callback Parameter | Description
 `data.sound` | `String` The name of the sound file to be played upon receipt of the notification.
 `data.image (windows only)` | `String` The path of the image file to be displayed in the notification.
 `data.additionalData` | `JSON Object` An optional collection of data sent by the 3rd party push service that does not fit in the above properties.
+`data.additionalData.foreground` | `Boolean` Whether the notification was received while the app was in the foreground
 
 #### Example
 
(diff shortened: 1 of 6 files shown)


Deminetix:
Unfortunately i based the branch of the head of my fork master, so fixes to previous issues are also in here. Not sure what i can do about it?

Deminetix:
Ideally we just need the last commit 9bf8ff6

Deminetix:
Closed as it has contains other issue branch's. I'll re-create the PR

Muji86:
So how do I specify the icon? do I just add an icon attribute like below?
What is the relative path to this? is it the project www/res folder?

 onDeviceReady: function () {
            "android": {
                "senderID": "1234567890987654321",
                "icon": "res/drawables/icon.png"
            },
            "ios": {"alert": "true", "badge": "true", "sound": "true"},
            "windows": {}
        });

Thanks

PR #46, opened 2015-07-16
add support for android icons/large-icons plus ability to override with the gcm keys

Deminetix:
This PR fixes the android lollipop issue displaying blank icons in issue #20 in a similar manner to what @macdonst mentioned

It adds support for both small-icon and large-icon, and gracefully falls back to how the plugin works prior to this PR if you don't set the android options

Small Icons:
- add the icon key to your android options which should be a string name of the resource to use (in res/drawables etc)
- if no android option is set, it falls back to what it currently does without this PR
- you can also override the android option key from the server by sending an icon key to GCM

Large Icons:
- add the largeIcon key to your android options which should be a string name of the resource to use (in res/drawables etc)
- if no android option is set, it does nothing
- you can also override the android option key from the server by sending a largeIcon key to GCM

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ Parameter | Description
 `options` | `JSON Object` platform specific initialization options.
 `options.android` | `JSON Object` Android specific initialization options.
 `options.android.senderID` | `String` Maps to the project number in the Google Developer Console.
+`options.android.icon` | `String` Optional. The name of a drawable resource to use as the small-icon. You can also override this option server-side by sending a `icon` key on the gcm data.
+`options.android.largeIcon` | `String` Optional. The name of a drawable resource to use as the large-icon. You can also override this option server-side by sending a `largeIcon` key on the gcm data.
 `options.ios` | `JSON Object` iOS specific initialization options.
 `options.windows` | `JSON Object` Windows specific initialization options.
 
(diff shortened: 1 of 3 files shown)


Deminetix:
Also added the plugin android option `iconColor` which sets the background color of the icon (see http://developer.android.com/reference/android/app/Notification.Builder.html#setColor(int))

macdonst:
@Deminetix I'm not going to merge this pull request but I am using it as the basis to add icon support to Android. Check out https://github.com/phonegap/phonegap-plugin-push/tree/icons

Basically, following your lead and allowing users to specify a drawable for the small icon and it's background color. Not allowing it to be set via GCM.

The large icon will be set from the "image" property that comes down from GCM. It checks for remote images over http, then android assets, then res/drawables before giving up and using the app icon.

Let me know what you think.

jayeshjain24:
@macdonst The large icon works seamlessly with drawables when pushed from GCM. But I am tired of testing my app again and again. I have tried literally everything from the docs that are mentioned here. Passed in the req params at initialization. Given drawables path,full url as `icon` but it just does not work. 

But I still dont know how to solve this problem. 

- remove small icon completely
- provide custom small icon
- show status bar icon instead of blank square.

status-bar
notification-tray

I moved from a pushplugin which was deprecated and dont wanna move to another plugin. can you please help?

macdonst:
@jayeshjain24 can you post your small icon file here?

jayeshjain24:
```
android: {
                    "senderID": "123456789"
                    "icon": "http://minio.co/images/jayeshjain.jpg"
                }

android: {
                    "senderID": "123456789"
                    "icon": "icon" //refers to icon in drawables.
                }
```

icon 

I just picked up an image from drawable-mdpi folder. I used ionic resources to crop icons and splashscreen which deposited few drawables folder. I didnt create any custom image. Hope I am clear

PR #59, opened 2015-08-10
Android sound fix

ericb:
The PushPlugin sets a `sound` property when the GCM message contains `sound` or `soundname`, but the GCMIntentService was still looking for the `soundname` property.


macdonst:
@ericb I made a fix to support both sound and soundname so we don't need to merge this anymore.

ericb:
Great, thank you!

PR #89, opened 2015-08-31
Android option `updateNotifications`

Deathspike:
Upon receiving a notification message, the current behavior will update
the existing notification message with the new content. In some cases,
this is not the desired behavior, which is where `updateNotifications`
comes in. By setting it to to `false` (from the default `true`), each
new message will produce a new notification message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Parameter | Description
 `options.android.sound` | `Boolean` Optional. If `true` it plays the sound specified in the push data or the default system sound. Default is `true`.
 `options.android.vibrate` | `Boolean` Optional. If `true` the device vibrates on receipt of notification. Default is `true`.
 `options.android.clearNotifications` | `Boolean` Optional. If `true` the app clears all pending notifications when it is closed. Default is `true`.
+`options.android.updateNotifications` | `Boolean` Optional. If `true` the app updates the existing notification when available. Default is `true`.
 `options.ios` | `JSON Object` iOS specific initialization options.
 `options.windows` | `JSON Object` Windows specific initialization options.
 
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -41,7 +41,8 @@ public class GCMIntentService extends GCMBaseIntentService {
     private static final String STYLE_PICTURE = "picture";
     private static final String STYLE_TEXT = "text";
     private static ArrayList<String> messageList = new ArrayList();
-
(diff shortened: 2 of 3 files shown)


Deathspike:
Just an additional comment: This is useful when you don't send notification identifiers from your push server (that is `notId`). For example, when you have multiple servers that aren't synchronized in any particular way but still supply notifications to devices. This implementation relieves your server(s) from having to track and synchronize these identifiers, and brings the entire behavior more in line with the expectations you'd have after dealing with APNS or WNS.

eros2187:
Hello Sorry you CAN post a code example because the Folder exsample there would Profit Thanks

Deathspike:
I'm sorry @eros2187, I don't really understand what an example would provide? It's an option, for Android, so you'd place it in the `init` call and it does exactly as I said; when you don't use a `notId` (that is, notification identifier) and set this option to `false`, it will generate its own notification identifiers which causes notifications to avoid overwriting each other. Can you elaborate what kind of example you want to see?

eros2187:
Thanks for your reply I used the old plug wax dove this Code Example
https://github.com/phonegap-build/PushPlugin/tree/master/Example/www

I could use an example of index.html because with the new plug - in does not work with

greetings eros2187

eros2187:
Can you help me ?

macdonst:
I'm not sure about this one, the Android way of handling this seems to be via 'tags'. Check out https://developers.google.com/cloud-messaging/server-ref#notification-payload-support as I'm thinking that this is the proper way to go.

Deathspike:
I agree that it would be better to follow the specifications, which states the following: "*If not set, each request creates a new notification.*". That would mean that the current `phonegap-plugin-push` implementation does not adhere to the specification and would require a breaking change going forward to correct the current behavior. Since I presume it to be extremely unlikely for a random contribution to introduce a breaking change, I opted for a similar-effect with an option without affecting the current behavior. But, by all means, I'm completely in favor in making it work as expected out-of-the-box.

macdonst:
@Deathspike yeah, my plan is to move from `notId` to `tags` as soon as I can. Once I do that it'll be a breaking change so I'll bump the version to 2.0.0. Feel free to take a stab at it yourself as PR's are welcome.

eros2187:
I would ask a help push the image of that size must have to be displayed to best effect?

thanks regards eros2187

mohitmayank:
@macdonst instead of moving from notId to tags. we can probably keep them both for android. if a tag is not provided in the push payload, we can use appName as the tag and keep the notId for tag level stacking.

Issue #105, opened 2015-09-05
Enhancement - Led, Vibration Pattern, Priority on Android


alexislg2:
Hey,
I could write a PR for supporting following:
- send a Led Color
- send a vibration pattern
- send priority (from 0 to 2)

For Android only of course. Interested?

macdonst:
@alexislg2 Yes, please do.

macdonst:
I'm reopening until I get a chance to merge your PR.

macdonst:
@alexislg2 re-test with the upcoming 1.3.0 release and re-open if it doesn't fix your issue.

ezequielrozen:
Hi, I'm making a push notification with node-gcm, and I don't know how to turn on the led in the phone. I tried as you showed in the example but it didn't work. Please can you tell me the way?

Thanks!

fredgalvao:
What device are you testing on?

Direct quote from android docs on [notificationBuilder.setLights](http://developer.android.com/reference/android/app/Notification.Builder.html#setLights(int, int, int)):
> Not all devices will honor all (or even any) of these values.

ezequielrozen:
@fredgalvao I just saw that there is a new release with the fix, but now I'm having this problem: https://github.com/phonegap/phonegap-plugin-push/issues/159

PR #108, opened 2015-09-06
ledColor, vibrationPattern and priority fields

alexislg2:
As announced here https://github.com/phonegap/phonegap-plugin-push/issues/105

I also removed all the trailing spaces on the Readme, hope you don't mind!


macdonst:
@alexislg2 thanks for the PR. I had to tweak a few things but it is not merged and will be in the next released version.

PR #133, opened 2015-09-14
Updating plugin.xml

fredgalvao:
Updating support.v13 to latest version and copying the new PushConstants.java over to platforms.


macdonst:
@fredgalvao holy cats! Thanks for that, I can't believe I forgot to add the new src file.

fredgalvao:
We all do, from time to time :smile:

PR #140, opened 2015-09-15
Bigview in inbox when only one message

alexislg2:
Currently when there is only one message in the inbox and that message is quite long we can only see the begginning of the message.

With this PR the user can expand the notification by unwraping it and can read the full message


macdonst:
@alexislg2 legit, I'll merge as soon as I can run my test.

PR #178, opened 2015-09-29
allow forceShow option to always notify (android only atm)

### LINKED ISSUE #76: Show notification even application is in foreground
It would be useful to have an option to show notification not only when application is in background or not running at all.
For example, to draw user's attention to events in familiar way without inventing disturbing message boxes or popups in application.

AdriVanHoudt:
fixes #76 

(not sure if done completely right though :D)

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -91,22 +91,36 @@ public class GCMIntentService extends GCMBaseIntentService implements PushConsta
             SharedPreferences prefs = context.getSharedPreferences(PushPlugin.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
             boolean forceShow = prefs.getBoolean(FORCE_SHOW, false);
 
-            // if we are in the foreground, just surface the payload, else post it to the statusbar
+            // if we are in the foreground and forceShow is `false` only send data
             if (!forceShow && PushPlugin.isInForeground()) {
                 extras.putBoolean(FOREGROUND, true);
                 PushPlugin.sendExtras(extras);
             }
+            // if we are in the foreground and forceShow is `true`, force show the notification if the data has at least a message or title
+            else if (forceShow && PushPlugin.isInForeground()) {
+                extras.putBoolean(FOREGROUND, true);
+                
+                showNotificationIfPossible(context, extras);
+            }
+            // if we are not in the foreground always send notification if the data has at least a message or title
             else {
                 extras.putBoolean(FOREGROUND, false);
 
-                // Send a notification if there is a message
-                String message = this.getMessageText(extras);
-                String title = getString(extras, TITLE, "");
-                if ((message != null && message.length() != 0) ||
-                        (title != null && title.length() != 0)) {
-                    createNotification(context, extras);
-                }
+                showNotificationIfPossible(context, extras);
             }
+        }
+    }
+    
+    private void showNotificationIfPossible (Context context, Bundle extras) {
+
+        // Send a notification if there is a message or title, otherwise just send data
+        String message = this.getMessageText(extras);
+        String title = getString(extras, TITLE, "");
+        if ((message != null && message.length() != 0) ||
+                (title != null && title.length() != 0)) {
+            createNotification(context, extras);
+        } else {
+            PushPlugin.sendExtras(extras);
         }
     }
 
@@ -214,7 +228,7 @@ public class GCMIntentService extends GCMBaseIntentService implements PushConsta
         setNotificationCount(extras, mBuilder);
 
         /*
-         * Notication add actions
+         * Notification add actions
          */
         createActions(extras, mBuilder, resources, packageName);
 
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -88,7 +88,7 @@ public class GCMIntentService extends GCMBaseIntentService implements PushConsta
         Bundle extras = intent.getExtras();
         if (extras != null) {
             // if we are in the foreground, just surface the payload, else post it to the statusbar
-            if (PushPlugin.isInForeground()) {
+            if (!FORCE_SHOW && PushPlugin.isInForeground()) {
                 extras.putBoolean(FOREGROUND, true);
                 PushPlugin.sendExtras(extras);
             }
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -87,8 +87,12 @@ public class GCMIntentService extends GCMBaseIntentService implements PushConsta
         // Extract the payload from the message
         Bundle extras = intent.getExtras();
         if (extras != null) {
+
+            SharedPreferences prefs = context.getSharedPreferences(PushPlugin.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
+            boolean forceShow = prefs.getBoolean(FORCE_SHOW, false);
+
             // if we are in the foreground, just surface the payload, else post it to the statusbar
-            if (!FORCE_SHOW && PushPlugin.isInForeground()) {
+            if (!forceShow && PushPlugin.isInForeground()) {
                 extras.putBoolean(FOREGROUND, true);
                 PushPlugin.sendExtras(extras);
             }
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Parameter | Description
 `options.android.sound` | `Boolean` Optional. If `true` it plays the sound specified in the push data or the default system sound. Default is `true`.
 `options.android.vibrate` | `Boolean` Optional. If `true` the device vibrates on receipt of notification. Default is `true`.
 `options.android.clearNotifications` | `Boolean` Optional. If `true` the app clears all pending notifications when it is closed. Default is `true`.
+`options.android.forceShow` | `Boolean` Optional. If `true` will always show a notification, even when the app is on the foreground. Default is `false`.
 `options.ios` | `JSON Object` iOS specific initialization options.
 `options.ios.alert` | `Boolean` Optional. If `true` the device shows an alert on receipt of notification. Default is `false`.
 `options.ios.badge` | `Boolean` Optional. If `true` the device sets the badge number on receipt of notification. Default is `false`.
--- a/src/android/com/adobe/phonegap/push/PushConstants.java
+++ b/src/android/com/adobe/phonegap/push/PushConstants.java
@@ -41,4 +41,5 @@ public interface PushConstants {
     public static final String COUNT = "count";
     public static final String FROM = "from";
     public static final String COLLAPSE_KEY = "collapse_key";
+    public static final String FORCE_SHOW = "forceShow";
 }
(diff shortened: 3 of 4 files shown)


src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -88,7 +88,7 @@ protected void onMessage(Context context, Intent intent) {
         Bundle extras = intent.getExtras();
         if (extras != null) {
             // if we are in the foreground, just surface the payload, else post it to the statusbar
-            if (PushPlugin.isInForeground()) {
+            if (!FORCE_SHOW && PushPlugin.isInForeground()) {
```
fredgalvao: Shouldn't you be getting this option in a way similar to this?
```java
SharedPreferences prefs = context.getSharedPreferences(PushPlugin.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
boolean force_show = prefs.getBoolean(FORCE_SHOW, false);
```

Accessing FORCE_SHOW directly is only targeting the `String` defined in PushConstants, and a `String` is always true'ish, thus making it not an option to decide on.

AdriVanHoudt: looks like you are right ^^ give me a minute to update

macdonst:
@AdriVanHoudt can you add the necessary documentation to the README.md file and then rebase so the PR merges cleanly? They I should be able to merge this.

AdriVanHoudt:
Will do

AdriVanHoudt:
Done

macdonst:
@AdriVanHoudt actually I did some testing of this today and it's not quite ready for merging. Take a look at my comments in the code.

AdriVanHoudt:
@macdonst did you add comments or do you mean the inline ones? Because I don't see any added on github atm

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -87,8 +87,12 @@ protected void onMessage(Context context, Intent intent) {
         // Extract the payload from the message
         Bundle extras = intent.getExtras();
         if (extras != null) {
+
+            SharedPreferences prefs = context.getSharedPreferences(PushPlugin.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
+            boolean forceShow = prefs.getBoolean(FORCE_SHOW, false);
+
             // if we are in the foreground, just surface the payload, else post it to the statusbar
-            if (PushPlugin.isInForeground()) {
+            if (!forceShow && PushPlugin.isInForeground()) {
```
macdonst: Since we have two variables we need to account for 4 different cases. 

1) forceShow = false & PushPlugin.isInForeground() = true
- should set foreground boolean to true and then fire the `notification` event.

2) forceShow = true & PushPlugin.isInForeground() = true
- should set foreground boolean to true, fire the `notification` event and show the notification in the shade

3) forceShow = false & PushPlugin.isInForeground() = false
- should set foreground boolean to false and then show the notification in the shade

4) forceShow = true & PushPlugin.isInForeground() = false
- should set foreground boolean to false and then show the notification in the shade

Obviously 3 and 4 are the same so we can handle that in an else but we need something like this un-optimized code:

```
            if (!forceShow && PushPlugin.isInForeground()) {
                extras.putBoolean(FOREGROUND, true);
                PushPlugin.sendExtras(extras);
            } else if (forceShow && PushPlugin.isInForeground()) {
                extras.putBoolean(FOREGROUND, true);
                PushPlugin.sendExtras(extras);

                // Send a notification if there is a message
                String message = this.getMessageText(extras);
                String title = getString(extras, TITLE, "");
                if ((message != null && message.length() != 0) ||
                        (title != null && title.length() != 0)) {
                    createNotification(context, extras);
                }
            } else {
                extras.putBoolean(FOREGROUND, false);

                // Send a notification if there is a message
                String message = this.getMessageText(extras);
                String title = getString(extras, TITLE, "");
                if ((message != null && message.length() != 0) ||
                        (title != null && title.length() != 0)) {
                    createNotification(context, extras);
                }
            }
```

macdonst:
@AdriVanHoudt just finished adding them. 

Also, when I had my OnePlusOne with the app running with your code I never got a notification event via JS or anything in the notification shade.

src/android/com/adobe/phonegap/push/GCMIntentService.java
fredgalvao: Doesn't case **2)** cause redundant behaviour? The `notification` event would be possibly triggered twice (one by the receipt of the message, and another by the click handler on the notificaton itself). I'm not saying that's wrong, I just think something else is needed to differ from both events, as both `data` objects will have the same properties as of right now.

AdriVanHoudt: Case 2 is wrong imo, the notification event should only be fired when actually clicked on the notification, as @fredgalvao said this could cause double events atm

macdonst: I see what you both mean. Remember that was vaguely pseudo code I posted. I still think we'd need to take care of case 2 when the information pushed to the device does not include a title or message parameter. Perhaps the developer wants to silently push some info to the device and still trigger the notification event. So more like:

```
            if (!forceShow && PushPlugin.isInForeground()) {
                extras.putBoolean(FOREGROUND, true);
                PushPlugin.sendExtras(extras);
            } else if (forceShow && PushPlugin.isInForeground()) {
                extras.putBoolean(FOREGROUND, true);

                // Send a notification if there is a message
                String message = this.getMessageText(extras);
                String title = getString(extras, TITLE, "");
                if ((message != null && message.length() != 0) ||
                        (title != null && title.length() != 0)) {
                    createNotification(context, extras);
                } else {
                    PushPlugin.sendExtras(extras);
                }
            } else {
                extras.putBoolean(FOREGROUND, false);

                // Send a notification if there is a message
                String message = this.getMessageText(extras);
                String title = getString(extras, TITLE, "");
                if ((message != null && message.length() != 0) ||
                        (title != null && title.length() != 0)) {
                    createNotification(context, extras);
                }
            }
```

fredgalvao: That fixes it well enough for me, but not for all cases... Well, if anything the user can always rely on the notId and check for duplicate events, I guess.

AdriVanHoudt: That works, my train of thought was that when there is an actual notification shown don't fire the event but, forceShow would not affect this when there is nothing to show
@fredgalvao what cases aren't covered?

fredgalvao: I'm not sure of the practical cases, but in theory one could send a push notification with `message === null` and still want it to be shown with something else (like a big picture, or a summaryText, or whatever else he might provide other than message).

AdriVanHoudt: If I am correct at least a message is required to show any notification

fredgalvao: You're right, the snippet @macdonst provided already contains the "message OR title" piece.

AdriVanHoudt: Will update today to match these comments

AdriVanHoudt:
Ok updated, only thing different then then the last code @macdonst proposed is that when not in foreground and no title/message it will send the extra data, which it doesn't do now

chriseckman:
Hey guys, I'm trying to get this to work, however the forceShow option doesn't seem to be working. I've tried to pull down the most recent changes, but it doesn't look like it's working on my N6.

AdriVanHoudt:
It is working for me when pulling directly from github in my config.xml `	<plugin name="phonegap-plugin-push" spec="https://github.com/phonegap/phonegap-plugin-push" />`
Also @macdonst any eta on a release, I don't like having to pull from github directly, maybe do a 1.3.1 or 1.4.0?

chriseckman:
@AdriVanHoudt thank you so much, your a god send. I had to add that into the config.xml to get it working. For whatever reason, "cordova plugin add https://github.com/phonegap/phonegap-plugin-push" wasn't pulling it in. 

In a side related question, what's the best way of displaying a confirmation popup if the app is in the foreground? I've found that $ionicPopup and confirm() doesn't work, the popup never displays. I was trying to ask the user if they would like to be redirected after receiving the notification, but it doesn't seem like anything is working. So for now, my work-a-around if to just always show a notification even if the device is in the foreground, then if they click the notification it redirects them. Any thoughts?

AdriVanHoudt:
Np :D I do not recoomend it though, why I asked for an eta on a release. 

And implementation depends on the app but you do know the notification event only fires when the notification is actually being clicked right? Unless it is only data or no forceShow and in foreground.

chriseckman:
@AdriVanHoudt Yes, that's the behavior I was aiming for. I was trying to only redirect to the new view if the user wanted to get to that view. The issue that I'm running into is that confirm() and $ionicPopup aren't displaying a popup when foreground === true, alert works fine so I know it's firing, but I'm not sure what the issue is with why confirm or $ionicPopup don't work. I was wondering if anyone had similar issues or had found something else that works. I was thinking the only other way is to use $watch to fire the $ionicPopup after I change the value of some $rootScope variable.

AdriVanHoudt:
@chriseckman I have no experience with $ionicPopup sorry :/ I redirect to a certain state on notification click and that works

macdonst:
@AdriVanHoudt I'm going to get a 1.4.0 out this month but an exact day is hard to pick right now as I'm at and off site this week and Cordova Face to Face meeting next week.

AdriVanHoudt:
Maybe do a prerelease? I don't want to set my dependency to the GitHub URL that fetches the master branch

AdriVanHoudt:
@macdonst sorry to bug you but any eta on a release? Just can't wait for the new features ^^

chriseckman:
Hey guys, another question. Finally got around to working setting up the iOS app with APN and I'm wondering if any has a workaround for the notifications not displaying if the iOS app is in the foreground?  Is there another plugin that would work for iOS?

AdriVanHoudt:
you can probably add an option for this as well (not really sure though), see https://github.com/phonegap/phonegap-plugin-push/blob/master/src/ios/AppDelegate%2Bnotification.m#L65 (just a guess though)

macdonst:
@AdriVanHoudt releasing today.

AdriVanHoudt:
:tada:

AdriVanHoudt:
@macdonst I think you missed this PR in the changelog ;)

luiscvalmeida:
@AdriVanHoudt does `forceShow` option blocks the `push.on('notification')` event from being called?
I just commented in here: #642

diegoherrera:
my app not Show notification when application is in foreground, could help me find the problem.

**my code in server is por example **

var registrationIds = [];
                        var sender = new gcm.Sender('xxxxxxxxxxxxxxxxxxxxxx');
                        var message = new gcm.Message();
                        message.addData('message',"u270C Peace, Love u2764 and PhoneGap u2706!");
                        message.addData('title','Push Notification Sample' );
                        message.addData('msgcnt','3');
                        message.addData('forceShow',true);
                        message.addData('additionalData',"{ id:1, name:'diego herrera' }");
                        message.addData('style', 'picture');
                        message.addData('picture', 'http://36.media.tumblr.com/c066cc2238103856c9ac506faa6f3bc2/tumblr_nmstmqtuo81tssmyno1_1280.jpg');
                        message.addData('summaryText', 'The internet is built on cat pictures');
                        message.timeToLive = 3000;
                        registrationIds.push('A...................................ccccccccccccccccc');
                        sender.send(message, registrationIds, 4, function (err,response) {
                            if(err) {
                                console.log("error " + err);
                            } else {
                                console.log(response);
                            }
                            });

**and my code in aplicattion is **

            "android": {
                "senderID": "xxxxxxxxxxx",
                "forceShow": "true"
            },
            "ios": {"alert": "true", "badge": "true", "sound": "true"}, 
            "windows": {} 
        });

         push.on('registration', function(data) {
            console.log("registration event");           
            console.log(JSON.stringify(data));
            app.gcm = data.registrationId; 
        });

         push.on('notification', function(data) {
            console.log("notification event");
            console.log(JSON.stringify(data));     
        });

         push.on('error', function(e) {
            console.log("push error");
        });

PR #182, opened 2015-09-30
Support the notification coming from Parse.com when the background

Hikaru-Ito:
Such as the push notification request from Parse.com, usually with different types of data will be sent.
Unfortunately, the current of this plug is not able to correspond to the format.
So, we added a program so as to correspond to Parce.com.


macdonst:
@Hikaru-Ito I appreciate this PR but I may not merge it. I need to look at way in which I can support different push data formats without having to write a bunch of if/else code.

macdonst:
@Hikaru-Ito thanks for the PR, I'm not going to merge it as #220 takes care of all the possible properties not just message. Please consider contributing in the future.

Hikaru-Ito:
ok. thanks!

PR #191, opened 2015-10-01
(doc) ios options

### LINKED ISSUE #180: iOS init option type
So, all examples and code samples call `.init()` with string options passed on the iOS part, but they're specified as Boolean on README.md. Checking source code I can see that both ways are handled, but shouldn't we either
* pick one way and go with it
or
* be explicit that both ways work
?

fredgalvao:
Being explicit about possible types/values on ios options: booleans can be passed as strings.
fixes #180


macdonst:
@fredgalvao I merged this and I have invited you to have write access so that you can just makes changes to the repo without having to do a PR. You've been a great help on this plugin.

fredgalvao:
It's been my pleasure to help! Thank you for the trust.

PR #220, opened 2015-10-13
Refactor key extraction into single method.

### LINKED ISSUE #147: Using this Plugin with Parse.com results in not showing Push on Android
I am experiencing a problem with Parse.com and iOS / Android Push-Plugin.
What I am doing is: sending from parse.com a push notification to iOS and Android devices with this Backend-command:

Parse.Push.send({
  		where: query, // Set our Installation query
 		data: {
    		alert: message
  		}
	}

So the type of this Notification is "alert". On iOS it is showing with no problem on Lockscreen or as Banner. But not on Android. If App is in Background I receive this on Console:

V/GCMBroadcastReceiver﹕ onReceive: com.google.android.c2dm.intent.RECEIVE
V/GCMBroadcastReceiver﹕ GCM IntentService class: com.adobe.phonegap.push.GCMIntentService
V/GCMBaseIntentService﹕ Acquiring wakelock
V/GCMBaseIntentService﹕ Intent service name: GCMIntentService-GCMIntentService-15
D/PushPlugin_GCMIntentService﹕ onMessage - context: android.app.Application@41bcf068
V/GCMBaseIntentService﹕ Releasing wakelock
(Related to this Issue: #86)

But nothing is showing up in Notification Center. As well I tried type "message" (as stated in #86) or "data" but nothing else happened. But if Android App is open I can receive type "alert" and show it via "alert(data.additionalData.data.alert)". Is this normal behavior? Because I'd like to have with the same command from my backend that the push is showing on android and ios. While I tried "message" and "data" it didn't show up on ios either. So how can I make Android showing up type "alert" in Notification Center and as Banner? Thanks!

smdvdsn:
This change moves all alternative key logic into a single method.
Allows message keys to be in nested in "data" or "message" object. This is to support parse.com and other providers.
Added additional keys to support Urban Airship.

fixes #147 replaces #218 and #182

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -31,6 +31,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Random;
 
 @SuppressLint("NewApi")
@@ -62,6 +63,8 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
             SharedPreferences prefs = getApplicationContext().getSharedPreferences(PushPlugin.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
             boolean forceShow = prefs.getBoolean(FORCE_SHOW, false);
 
+            normalizeExtras(extras);
+
             // if we are in the foreground and forceShow is `false` only send data
             if (!forceShow && PushPlugin.isInForeground()) {
                 extras.putBoolean(FOREGROUND, true);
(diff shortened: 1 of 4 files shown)


smdvdsn:
Right now this code supports all the providers we know about but I still think we should allow the keys to be externally configurable. I wanted your input on how best to do that. I'm thinking we support the simple alternate keys by default and then allow additional configuration to be passed in to support providers like parse.com and Urban Airship etc.

The config would include some kind of map that the nomalizeKey method could use, a list of prefixes to strip and a list of keys for nested objects to extract.

```
{
  "alternateKeys": {
      "message": ["message", "alert"],
      "count": ["msgcount", "badge"],
      ....
  },
  "stripPrefix": ["gcm.notification", "com.urbanairship.push"],
  "alternatePayload": ["data", "message"]
}
```

macdonst:
@smdvdsn Yup, I'm going to merge this now. I agree with your assessment that there should be a configuration file to be able to support other push services. Let me know if you would like committer access to the repo and we can work this out on another branch.

smdvdsn:
@macdonst I'm happy to keep working on this on another branch. I'm wondering if there is anyway to make this a build time config?

macdonst:
@smdvdsn I think run-time config makes more sense. I'll create a new branch today.

PR #232, opened 2015-10-15
Add getApplicationIconBadgeNumber method

rubenstolk:


macdonst:
Seems legit but can you update the readme with instructions on the new method.

rubenstolk:
@macdonst I think you were a bit faster than me because I was still doing that :) Amended the commit

macdonst:
@rubenstolk lemme know when you are done.

rubenstolk:
@macdonst It's done now.

macdonst:
@rubenstolk great stuff, thanks.

Issue #188, opened 2015-10-01
Use Google's InstanceID API


macdonst:
https://developers.google.com/instance-id/guides/android-implementation

bhaveshgohel:
:+1:

bau720123:
if could
well..that will really awesome!!!

miraculixx:
To add some perspective -- this is not an _awesome_ feature but a requirement by Google:

According to [this](https://developers.google.com/cloud-messaging/android/legacy-regid), the `gcm.register` is deprecated since May 2015:

_GCM register() is deprecated starting May 28, 2015. New app development should use the Instance ID API to handle the creation, rotation, and updating of registration tokens._

While there do not seem to be direct implications (it still works), it would be prudent to plan for migration for the day will come when GCM stops accepting new register() calls.

macdonst:
@miraculixx Agreed, that's why I'm working on it. :smiley_cat:

miraculixx:
@macdonst glad to hear :+1: thanks

macdonst:
Anyone who wants to test it can check out branch `issue188`. It seems to be working just fine but I need to do more testing.

macdonst:
Fix committed will be official in 1.4.0 release.

bhaveshgohel:
Thanks @macdonst ! :+1: 
Can you please also update the README.md file to reflect the changes on how to use it. :smiley:

bau720123:
+1

macdonst:
@bhaveshgohel I already did under Android compilation. No changes to JS code are necessary.

bhaveshgohel:
Ohh... Thanks @macdonst! :smiley:

jayj:
It added the "Find accounts on the device" Contact permission. What is that permission used for, if it's used?

macdonst:
@jayj Hmm...that is Google Play Services adding that. Lemme see if I can make that smaller.

macdonst:
@jayj I just merged a PR from contributor @smdvdsn that should remove the need for the "Find accounts on device" permission.

chrillewoodz:
@macdonst  I've looked everywhere but I cannot find how to get hold of an InstanceID using this plugin... Can you shed some light on it?

PR #235, opened 2015-10-18
Fixes for using new GcmReceiver code

smdvdsn:
Replace intent category "com.example.gcm" with $PACKAGE_NAME in plugin.xml.
Change play services dependancy to more specific play-services-gcm

update issue #188


macdonst:
@smdvdsn thanks, you rock

PR #240, opened 2015-10-19
added notification cancel function

mohitmayank:
Added ability to cancel a notification in android using notId.


macdonst:
@mohitmayank have you tested what happens if you call cancel notification on an iOS device?

mohitmayank:
@macdonst no. i haven't. Don't have any ios device as of yet. but i plan to start ios app soon and will complete this feature for ios too. and windows as well.
as of now, maybe we can mention this in documentation that its only for android. 
there seems to be a way of cancelling notifications in ios http://stackoverflow.com/questions/6340664/delete-a-particular-local-notification

not sure about windows.

macdonst:
superseded by #252

PR #252, opened 2015-10-21
Android notification tags and cancel feature

mohitmayank:


mohitmayank:
if you merge this one, previous pull request https://github.com/phonegap/phonegap-plugin-push/pull/240 can be ignored as it contains changes from that and some more modifications. but this one will be a breaking change on the previous one.

center777:
after call cancel method, messageMap in class GCMIntentService will be contains list of inbox messages. Don't forget clear messageMap.

PR #305, opened 2015-11-04
check whether Push Notification Permissions have been enabled for this app

zwacky:
as soon as you run `var push = window.PushNotification.init({ ...` the app will ask the user for push notification permission. with this you have more flexibility.

example usage:

```
window.PushNotification.hasPermission((data) => {
                if (data.isEnabled) {
                    var push = window.PushNotification.init({
                        android: {
                            senderID: 'abc'
                        }
                    });
                }
```

it will always return a json in format:

```
{
    isEnabled: true|false
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This requires phonegap/cordova CLI 5.0+ ( current stable v1.4.2 )
 ```
 phonegap plugin add phonegap-plugin-push
 ```
-or 
+or
 
 ```
 cordova plugin add phonegap-plugin-push
@@ -21,7 +21,7 @@ It is also possible to install via repo url directly ( unstable )
 phonegap plugin add https://github.com/phonegap/phonegap-plugin-push
 ```
 
-or 
+or
 
 ```
 cordova plugin add https://github.com/phonegap/phonegap-plugin-push
@@ -202,6 +202,28 @@ successHandler gets called when background push processing is successfully compl
 push.finish(successHandler, errorHandler);
 ```
 
+### PushNotification.hasPermission(callback)
+
+Checks whether the PushNotification Permission have been enabled or disabled.
+
+Callback Parameter | Description
+------------------ | -----------
+`data.isEnabled` | `Boolean`
+
+#### Example
+
+```javascript
+PushNotification.hasPermission(function(data) {
+    if (data.isEnabled) {
+        var push = PushNotification.init({
+            android: {
+                senderId: 'abc'
+            }
+        });
+    }
+});
+```
+
 ## PhoneGap Build Support
 
 Including this plugin in a project that is built by PhoneGap Build is as easy as adding:
@@ -246,10 +268,10 @@ This is because Android now uses Material design and the default icon for push w
 In order to get a better user experience you can specify an alternate icon and background color to be shown when receiving a push notification. The code would look like this:
 
 ```javascript
-	var push = PushNotification.init({ 
-		"android": { 
-			"senderID": "123456789", "icon": "phonegap", "iconColor": "blue"}, 
-		"ios": {"alert": "true", "badge": "true", "sound": "true"}, "windows": {} 
+	var push = PushNotification.init({
+		"android": {
+			"senderID": "123456789", "icon": "phonegap", "iconColor": "blue"},
+		"ios": {"alert": "true", "badge": "true", "sound": "true"}, "windows": {}
 	});
 ```
 
@@ -482,7 +504,7 @@ To add to your app:
 ```
 phonegap plugin add https://github.com/jeduan/cordova-plugin-facebook4 --variable APP_ID="App ID" --variable APP_NAME="App Name"
 ```
-or 
+or
 
 ```
 cordova plugin add https://github.com/jeduan/cordova-plugin-facebook4 --variable APP_ID="App ID" --variable APP_NAME="App Name"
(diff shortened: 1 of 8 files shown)


macdonst:
@zwacky did a quick look and it seems fine. Can you add some explanation in the README file?

zwacky:
yes i will

zwacky:
@macdonst hope it's alright

fredgalvao:
Making a note here so that we remember to take a new look at this with the new permission stuff from android-5.0, things might be easier with it.

stinju:
@macdonst in reference to https://github.com/phonegap/phonegap-plugin-push/issues/326, this was the patch that I put in, that seemed to disrupt the notification handling, when used perhaps inappropriately.  I traced it down to the fact that in this patch, ```pushContext``` is set in the ```hasPermission``` logic.  I am using ```hasPermission``` as a way to detect permissions at any moment, so instead of using it as @zwacky described above, I was invoking it even after a ```window.PushNotification.init``` call.  When used in this way, ```init```'s ```pushContext``` is overridden by ```hasPermission```, and subsequent notification events went through the context (callback) provided by ```hasPermission``` instead of ```init```.

Given this, with regards to this patch, it might be wise to either 1) handle the ```hasPermission``` callback without setting ```pushContext```, 2) provide an error if ```hasPermission``` is called after ```init```, or 3) provide a note in the docs.  I have no expertise here, so I defer to your judgment, but I just wanted to give you a heads up.

zwacky:
@stinju i see what you mean. i definitely would expect it to be used after `init`.

so instead of using the `sendEvent(JSONObject _json)` in android, it should directly call `callbackContext.sendPluginResult(pluginResult)`...

gotta test things first.

macdonst:
@zwacky yeah, use the new callbackContext for that method call. The sendEvent method is a convenience method for all of the remote notifications that come in.

zwacky:
also found out, the the android *minSDK = 19* (android 4.4) since it uses `android.app.AppOpsManager`.

QuentinFchx:
What is blocking this PR to be merged ?

zwacky:
this should fix the last issue @stinju mentioned.

serratus:
@zwacky In your example you call `window.PushNotification.hasPermission` before calling `window.PushNotification.init`. What happens when the user opens the app for the first time? Does `hasPermission` also ask users if they want to receive notifications or not? I noticed, upon first load of the app, it does not have permission to send notifications, and therefore the `init` would never run?! Can you please clarify on that?

zwacky:
it's just an additional function, it that has nothing to do with `init()`.

it's just to prevent the user from having a push notification permission request popup at startup, that most probably will be declined.

you can then call `init()` at a very specific time, so more users would accept the request.

with `hasPermission()` you can `init()` upon every startup then to keep the tokens up to date.

serratus:
Thanks for the feedback. But my question was more focused on the fact that `init()` asks the user for permission if it happens for the first time. How do I know whether the user accepts the permission, or denies it? Is there any callback? Because I've noticed that the `registration` event is even fired when permission is denied. Do I have to poll the status afterwards with `hasPermission()`? Or am I missing something else here?

zwacky:
yep, polling the permission. that's how i track it though.

serratus:
Good to know, thanks for the clarification.

macdonst:
@zwacky Do you have any time to resolve the merge conflicts on this PR?

zwacky:
maybe i'm blind... where can i see the conflicts?

macdonst:
@zwacky Oh right, you don't see the Merge box because you don't have committer access. Leave it with me.

macdonst:
@zwacky I have a problem with the example:

```
window.PushNotification.hasPermission((data) => {
                if (data.isEnabled) {
                    var push = window.PushNotification.init({
                        android: {
                            senderID: 'abc'
                        }
                    });
                }
});
```

If I'm starting up a brand new app on iOS then `data.isEnabled` is always `false`. That's to be expected as the push permission dialog doesn't get show until `init` is called but in this case `init` will never get called. Thoughts?

macdonst:
@zwacky actually it is the same thing on Android as well. The permission check works great but we may want to change the example docs.

macdonst:
Merged into v1.5.x branch

zwacky:
yeah you're right. should have a hint, that you need to check for that `init` somewhere else, manually.

will update the docs.

macdonst:
@zwacky if you update the docs please do so in the v1.5.x branch which is the current dev branch.

zwacky:
i did: #425

macdonst:
I'm closing this PR as it has been merged into v1.5.x and will become part of master soon.

JeremyColton:
So, call init() first and then poll for a response using hasPermission()

But, hasPermission() returns false the first time the app is open since push has never been granted, so polling using this function will return false while X seconds later, the user may allow push notifications and then hasPermission() will return true.  So, what's the best practise here - poll for X seconds waiting for a true.  After X seconds, assume the user has answered the "allow push notiffication" popup and cancel the poll if it still replies false?

This seems wrong and inefficient! Any advice?

zwacky:
@JeremyColton there isn't a nice and sophisticated way than trying, because we don't have hooks for the native permission request modal.

i ask the user, if he'd like to stay up to date. then if he does, i `init()` (the native permission request modal shows up) and i start the polling 10x for 2s.

jaigtz88:
@JeremyColton , so do you have any solution to solve that issue?

PR #322, opened 2015-11-10
add sound default statement

LightZam:
Hi, there is a problem about android sound

ios will ring by setting sound as 'default' even app was not in background or foreground, 

but android will take 'default' value as a media file, 

i think android should provide 'default' value to indicate system default sound,

according to gcm document : https://developers.google.com/cloud-messaging/http-server-ref#table2

that is the reason for this pull request

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -75,7 +75,7 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
             else if (forceShow && PushPlugin.isInForeground()) {
                 Log.d(LOG_TAG, "foreground force");
                 extras.putBoolean(FOREGROUND, true);
-                
+
                 showNotificationIfPossible(getApplicationContext(), extras);
             }
             // if we are not in the foreground always send notification if the data has at least a message or title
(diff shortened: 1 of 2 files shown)


macdonst:
@LightZam yeah, I see what you mean. In order to get the behaviour you are looking for you only need to do:

```
var push = PushNotification.init({ "android": {"senderID": "12345679", sound: true},
} );
```

to tell Android to play the default sound when receiving a notification. Then your payload coming from GCM doesn't even need a `sound` property.

However, what you have sent has merit and I'll merge it soon.

fredgalvao:
@macdonst is that `sound:boolean` property on the "android" config already in place and functional? If so, we should mention it on the docs.

macdonst:
@fredgalvao yup, it is already in the options table.

https://github.com/phonegap/phonegap-plugin-push#pushnotificationinitoptions

LightZam:
i guess you are misunderstanding my describe,
i know about sound option, and i had set it as true value,
but when push coming like below
    
    {  
    "notification":{
      "title":"Zam",
      "body":"test23",
      "sound":"default",
      "image":"icon_push",
      "badge":5
    },
    "data":{
      "notId":"12112"
    }
    "registration_ids":[
      "fR9Z2TH_Pt4:APA91bF_mv68BC2J-2WS8zvdZILeiZdbvaGWXmZ_3ORVx3ltLkzOaCVvdKbvurpSgzAg7nR5jAEmEOgHLRYZnk5-1tZvZAPCXQM2ceE1J6OhhzJBb2iIiGQeSFQvJjnvVoKwous2nwwU"
    ]
    }

i used sound as default for ios required default

but when i use that, android sound was gone forever, because it can't find any media named sound.

but i guess gcm document give us a advice to use 'default' value as system default sound too

PR #323, opened 2015-11-11
Added event handler removal

giordanocardillo:


fredgalvao:
Thanks for your contribution!
The patch seems fine, but there is no documentation. Could you add it and a practical example?

macdonst:
and tests if possible.

giordanocardillo:
Sorry, I'm new here, how am I supposed to add tests and docs?

macdonst:
For docs just add the information in the README.md.

For tests if you have run `npm install` in the root of the phonegap-plugin-push directory it should install the npm packages required for running the tests. Then you just edit the `spec/index.spec.js` to add your tests.

giordanocardillo:
## PushNotification event handler removal

This function lets you remove an event handler from the plugin handlers collection.

### Usage

1. Initialize the PushNotification plugin
  ```javascript
//PushNotification plugin initialization
 var pn = PushNotification.init({ /*...*/ });
```

2. Define a function to handle the event
  ```javascript
var eventFunction = function(data){
  /*...*/
}
```

3. Add the event handler to the plugin, so the function gets called whenever the event is triggered
  ```javascript
//Adding event handler
pn.on('notification', eventFunction);
```
  Now everytime the app receives a notification, the `eventFunction` gets called.

4. If you want to remove that handler 'cause you don't need it anymore, simply call the following function
   ```javascript
//Removing event handler
pn.off('notification', eventFunction);
```

#### Note"
**You will have to pass the same `eventFunction` to match the exact handle to remove from the handlers collection.**

macdonst:
Seems good but add it to README.md and update your PR.

giordanocardillo:
OK, documentation added and pushed, now I'm getting on tests

giordanocardillo:
Tests inserted, you can check it

PR #359, opened 2015-11-19
Now calling .trim() to each array String element when parsing them to Ints/Longs

SandroGrzicic:
When a vibration pattern is sent as e.g. [0, 150, 250, 100], or a ledColor as e.g. [0, 255, 127, 64], the space after each comma will now be trimmed, which would otherwise cause a NumberFormatException when trying to parseInt/parseLong the string " 0255".


SandroGrzicic:
A workaround would be to simply send these arrays as e.g. [0,150,250,100] from the server.

macdonst:
Thanks @SandroGrzicic

SandroGrzicic:
Thank you for the great plugin.

PR #383, opened 2015-11-27
forceShow on ios

AdriVanHoudt:
* works :tada: 
* not sure what happened with the styling there(I blame xcode) but I will fix this no worries
* what is not working
    * no sound or vibration on `forceShow` when in foreground (ios thing?)
    * when app is open and notification from `forceShow `is made and you drag down the notifications center and then push it back up, `appDidBecomeActive` is called and the notification is 'executed'. I do not know enough cordova api or ios api stuff to fix this. :sos: 

* any suggestions for improvements welcome!

EDIT yes I will squash those commits when done :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/ios/AppDelegate+notification.m
+++ b/src/ios/AppDelegate+notification.m
@@ -33,8 +33,8 @@ static char launchNotificationKey;
 - (AppDelegate *)swizzled_init
 {
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(createNotificationChecker:)
-                name:@"UIApplicationDidFinishLaunchingNotification" object:nil];
-    
+               name:@"UIApplicationDidFinishLaunchingNotification" object:nil];
+
     // This actually calls the original init method over in AppDelegate. Equivilent to calling super
     // on an overrided method, this is not recursive, although it appears that way. neat huh?
     return [self swizzled_init];
--- a/src/ios/AppDelegate+notification.m
+++ b/src/ios/AppDelegate+notification.m
@@ -24,7 +24,7 @@ static char launchNotificationKey;
 + (void)load
 {
     Method original, swizzled;
-    
+
     original = class_getInstanceMethod(self, @selector(init));
     swizzled = class_getInstanceMethod(self, @selector(swizzled_init));
     method_exchangeImplementations(original, swizzled);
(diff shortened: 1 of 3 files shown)


vlaraort:
I tried your code and I dont understand the behaviour. It is intended to show the notification, in the notification bar, when the app is in foreground right? just like in Android.

AdriVanHoudt:
@vlaraort jup

vlaraort:
@AdriVanHoudt I am unable to make it work. It just calls the notificationHandler in the native code, but the notification doesn't appears. Looking in Apple documentation says:

"When the operating system delivers a local notification or remote notification and the target app is not running in the foreground, it can present the notification to the user through an alert, icon badge number, or sound.

If the app is running in the foreground when the notification is delivered, the app delegate receives a local or remote notification."

And looking throught stackoverflow, nobody are able to make it work, so they use custom alerts or popus instead. 

I tested in iPad2 with IOS 9.0

AdriVanHoudt:
could be that you need to pass `"content-available": 1` for this, can't remember

vlaraort:
I am passing it, content-available and forceShow in the initialization, and still not works

macdonst:
@AdriVanHoudt I finally had some time to merge this into my current dev branch and I'm seeing the same thing as @vlaraort. I don't think you can force a local notification to be shown like this. Have you looked at this code in awhile?

AdriVanHoudt:
I haven't touched it since then and I am no iOS dev so it is quite possible it is broken

paramsen:
There is

paramsen:
I'm afraid that here is *no* possibility for this to work on IOS without custom alerts/banners. From the docs:

"When your app is running but not frontmost, UIKit displays the system UI and then delivers the results to your app in the background when possible. Similar to when your app is running, UIKit calls methods of your app delegate to receive the notification data and perform any actions. If it is unable to deliver the notification data in the background, UIKit waits until the next time your app runs to deliver it."

This means that banners/alerts will *only* be displayed if the app is not in the foreground. Otherwise the OS will deliver the push directly to your app (AppDelegate:didReceiveNotification) which in Cordova/PushPlugin forwards the push to PushPlugin:on('notification')-listener.

macdonst:
@AdriVanHoudt I'm closing this PR as the approach will never work on iOS.

AdriVanHoudt:
np I hope this will become possible one day!

PR #389, opened 2015-12-01
Uses current flavor app id

CookieCookson:
Currently the plugin pulls the `PACKAGE_NAME`, early on in the build process. After working with Android product flavors where I am building different APKs with different applicationIds (com.test.app.alpha, com.test.app.beta, com.test.app) the permission still sticks with the core package name. These permission requests then conflict with each other when having multiple APKs on the same device, and can even stop the app from being downloaded from the Play Store if the user has an another app on the device with the same permissions.

By using `{applicationId}` it will grab the current App flavor's application ID later on so the current ID is used instead.

applicationId is fairly recent, so I am unsure about backwards compatibility with this fix. I've done 'some' testing too, but nothing thorough at this stage.


macdonst:
@CookieCookson good idea, thanks for the PR.

PR #397, opened 2015-12-04
Background notification should trigger event if possible

renanccastro:
Notification event should always be sent, if possible. This removes the problem where notification gets to the desk, but event handler is never called.

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -206,10 +206,11 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
             Log.d(LOG_TAG, "create notification");
 
             createNotification(context, extras);
-        } else {
-            Log.d(LOG_TAG, "send notification event");
-            PushPlugin.sendExtras(extras);
         }
+        //Should always send notification event
+        Log.d(LOG_TAG, "send notification event");
+        PushPlugin.sendExtras(extras);
+        
     }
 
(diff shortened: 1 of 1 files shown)


fredgalvao:
There's currently only one case where it doesn't immediately dispatch the `notification` event, which is when it has a title or a message, and in this case it creates a notification whose click will eventually dispatch that `notification` event as soon as the user interacts with the notification.

Although I see what you're trying to accomplish here, and I agree that it's a good idea (which is to somewhat have two different events, one for `notification.receive` and another for `notification.click`, maybe even a third one `notification.discard`), but I disagree this is the way to go.

With your patch, we would have multiple `notification` events for the same object (considering it has either title or message): 
- one when it arrives(the one you patched)
- and another one when the user clicks on it (the one already in place)

and that's kinda inconsistent.
We would at the very least have to document that quirky case.

macdonst:
@renanccastro sorry, while I appreciate the PR I'm not going to merge it. @fredgalvao provides a very detailed explanation as to why this is the case. The way this is going to be handled in the 1.5.0 release is if the notification has no title/message it will call the `on('notification')` event as it does now. However, if you add a `content-available: 1` property to the notification it will force the `on('notification')` handler to be called just like iOS.

fredgalvao:
@macdonst  Wouldn't all of these issues regarding `background`/`cold start`/`content-available`/`force-show` be solved if we just split the `notification` event into multiple ones like I mentioned? Each one of them would be easily configurable and wouldn't interfere with each other that much. Also, we'd have the path clear for the android part on `cold start` notifications having no callback.

PR #400, opened 2015-12-07
Implements an ArrayList<Bundle> for the extras

nicoabie:
To avoid loosing the cached previous extra when a new one arrives

### DIFF
--- a/src/android/com/adobe/phonegap/push/PushPlugin.java
+++ b/src/android/com/adobe/phonegap/push/PushPlugin.java
@@ -28,7 +28,7 @@ public class PushPlugin extends CordovaPlugin implements PushConstants {
 
     private static CallbackContext pushContext;
     private static CordovaWebView gWebView;
-    private static Bundle gCachedExtras = null;
+    private static ArrayList<Bundle> gCachedExtras = new ArrayList<Bundle>();
     private static boolean gForeground = false;
 
     /**
@@ -119,10 +119,13 @@ public class PushPlugin extends CordovaPlugin implements PushConstants {
                         editor.commit();
                     }
 
-                    if (gCachedExtras != null) {
+                    if (!gCachedExtras.isEmpty()) {
                         Log.v(LOG_TAG, "sending cached extras");
-                        sendExtras(gCachedExtras);
-                        gCachedExtras = null;
+                        Iterator<Bundle> gCachedExtrasIterator = gCachedExtras.iterator();
+                        while (gCachedExtrasIterator.hasNext()) {
+                            sendExtras(gCachedExtrasIterator.next());
+                        }
+                        gCachedExtras.clear();
                     }
                 }
             });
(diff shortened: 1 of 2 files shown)


src/android/com/adobe/phonegap/push/PushPlugin.java
```diff
@@ -119,10 +119,13 @@ else if (!savedSenderID.equals(senderID)) {
                         editor.commit();
                     }
 
-                    if (gCachedExtras != null) {
+                    if (!gCachedExtras.isEmpty()) {
                         Log.v(LOG_TAG, "sending cached extras");
-                        sendExtras(gCachedExtras);
-                        gCachedExtras = null;
+                        Iterator<Bundle> gCachedExtrasIterator = gCachedExtras.iterator();
```
macdonst: You are creating an Iterator here but there is a possible concurrency issue if a new push notification comes in while you are iterating through the current notifications. Probably best off to make a copy of `gCachedExtrasIterator` then empty `gCachedExtrasIterator` before you start iterating.

src/android/com/adobe/phonegap/push/PushPlugin.java
```diff
@@ -119,10 +119,13 @@ else if (!savedSenderID.equals(senderID)) {
...
+                        while (gCachedExtrasIterator.hasNext()) {
+                            sendExtras(gCachedExtrasIterator.next());
```
macdonst: Instead of sending multiple notification events have you thought of creating a new method `sendExtras(ArrayList<Bundle> )` that would end up sending one `notification` event with an array of notification data objects?

nicoabie: I have thought of it indeed but decided not to fall into a breaking change for now. I like the idea, are you OK with making that change into the API?

src/android/com/adobe/phonegap/push/PushPlugin.java
nicoabie: Oh yes, thanks for that. Will take a look into this :)

src/android/com/adobe/phonegap/push/PushPlugin.java
macdonst: I think it is cleaner and targeting this for a 1.5.0 release anyway so I'm okay with making that change.

pallaviMN:
Hi Team,

Can you please provide a fix for this? I am stuck from past 3 days in this issue.

mikepsinn:
Hi, guys!  

I just created this issue https://github.com/phonegap/phonegap-plugin-push/issues/1271 and then realized there's already a pull request with the fix.  

I'm not really an Android programmer so the comments are a little over my head, but it seems like the benefits of the merge (allowing multiple notifications with action buttons to work, at least sometimes) would exceed the risks. 

What do you think?

Thanks! :D

macdonst:
@mikepsinn this PR fixes a different issue but it doesn't quite work. Your #1271 issue needs more investigation as clicking an action button should start the app.

mikepsinn:
Got it. Thanks for the explanation, @macdonst !

macdonst:
Fixed by #1272

PR #409, opened 2015-12-11
Cordova plugin test framework

nicoabie:
I wanted to create a test for the data.additionalData callback argument. I wrote it and it obviously failed because all test where running on node where the is no DOM and no window object. :(
Then I tried to mock it with no luck, I did some research and found the [cordova-plugin-test-framework](https://github.com/apache/cordova-plugin-test-framework) and after some more research something to make it run [cordova-paramedic](https://github.com/purplecabbage/cordova-paramedic/blob/master/package.json)
I found some flaws on the cordova-paramedic runner like there is no way to reuse a previously created app just reinstalling the plugins and that there is no livereload. 
I would take a look maybe I can implement those features in the cordova-paramedic to make the cycle faster.

 @macdonst if you know any other way to run the cordova-plugin-test-framework I'm open to suggestions.

I think it would be nice to use it maybe this is not the best way yet.


macdonst:
@nicoabie I'm not going to accept this PR in it's current format. Please leave the spec tests like they are right now. If you want to run cordova-paramedic, which I encourage, then you should look at https://github.com/phonegap/phonegap-plugin-contentsync which includes both jasmine-node and paramedic tests.

tests/tests.js
```diff
@@ -0,0 +1,303 @@
+/* globals require, cordova */
+
+/*!
+ * Module dependencies.
+ */
+
+var PushNotification = require('phonegap-plugin-push.PushNotification'),
```
macdonst: Shouldn't need to add this line.

www/push.js
```diff
@@ -59,7 +53,7 @@ var PushNotification = function(options) {
 
     // wait at least one process tick to allow event subscriptions
     setTimeout(function() {
-        exec(success, fail, 'PushNotification', 'init', [options]);
+        cordova.exec(success, fail, 'PushNotification', 'init', [options]);
```
macdonst: Running in paramedic shouldn't require this to be pre-fixed with cordova.

nicoabie:
I'll look into it @macdonst, thanks for the review. Are you ok with including the cordova-plugin-test-framework? To be able to do test like [this](https://github.com/nicoabie/phonegap-plugin-push/commit/8e7d38800ac494a5fceb2dcfddba1771a960ebe5) using the window object.

www/push.js
nicoabie: I created a ```spyOn(cordova, 'exec')``` that's why I couldn't use at the top the ```var exec = require('cordova/exec')```, to make the spy work

Issue #333, opened 2015-11-13
app closed launch function??


eros2187:
hello I wanted to know if it was possible to launch a function of timing and make it run even if the app is closed ?
there is a plugin as it does wotzapp that remains always open to speak!!
thank you

nicoabie:
To launch it where? If the app is closed you don't have access to the webview. Could you elaborate?

EternallLight:
Hello,
I have probably a similar question. 
Is it possible to launch the app if it is closed on receiving a notification? I tried the high with notifications with high priority - unfortunately, it does not help.

nicoabie:
If the app is not running (the plugin's webview is null) it caches the notification payload data [see here](https://github.com/phonegap/phonegap-plugin-push/blob/master/src/android/com/adobe/phonegap/push/PushPlugin.java#L185)
When the user taps on the notification that launches the app and when you register the notifications callback you are going to receive the cached payload. Is that not working for you?

EternallLight:
Nicolás, thanks for your quick response.
Sure, it's working. However, in our case, we need to get the app running immediately without a notification (no 'title' or 'body' to specify in the payload) - say, for an incoming call, which is supposed to fire once the app gets loaded. I thought it's a default behaviour for all high priority notifications, but it seems that I was wrong.

nicoabie:
I think I see what you mean, but I don't know if that is possible check [this link](https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/ApplePushService.html) about content-available flag.
I'm not sure if you can put the app in foreground.
There is some work being done on android to emulate that behavior aswell [this issue](https://github.com/phonegap/phonegap-plugin-push/issues/293)

EternallLight:
From [this changes](https://github.com/phonegap/phonegap-plugin-push/commit/2b156daf2a2e0b9943fda03e1a6f52632e891394), I see that the behaviour with "content-available: 1" is the same as now with no "title" and "body" keys. Unfortunately, it does not instantly launch the application if it is closed - the notification gets delivered to JavaScript only on the next app run.

nicoabie:
Yes exactly, I don't know if it is possible to launch it without user interaction.

EternallLight:
Well, Skype somehow does that unless force closed. Now I'm trying to figure out how.

nicoabie:
True, keep us updated with your findings ;)

[...]

PR #471, opened 2016-01-05
Force running the application if it was closed (Issue #333)

EternallLight:
Hi,

We've made some improvements to achieve running the application if it was closed at the moment of GCM notification - with bringing the app to the foreground or starting it silently on the background. This works only on the Android platform. Please review the changes and include them in the project or let me know what I should improve/change to get them included. 

Thanks.

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -197,6 +197,7 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
         String message = extras.getString(MESSAGE);
         String title = extras.getString(TITLE);
         String forceLaunch = extras.getString(FORCE_LAUNCH);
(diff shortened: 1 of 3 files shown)


EternallLight:
Forgot to mention one improtant thing.
The silent run on the app requires changes the app's MainAcrtivity's onCreate method like the following:
...
Bundle extras = getIntent().getExtras();
if (extras != null) {
    String startOnBackground = extras.getString("start_on_background");
    if("1".equals(startOnBackground)) {
        Log.d(LOG_TAG, "Starting Main Activity on background...");
        moveTaskToBack(true);
    }
}

So if this is contrary to plugin's principles, please omit the second commit.

macdonst:
Closing as I'll be able to fix this with some new changes in cordova-android.

EternallLight:
Simon, could you kindly specify which exactly changes are related? I updated cordova-android and now see some unpredictable behavior like pause/resume events triggered several times.

PR #624, opened 2016-02-16
Add setApplicationIconBadgeNumber action for Android, using ShortcutBadger

### LINKED ISSUE #622: Android badge count
Android badge count is not supported by default, but is does appear to be somewhat standardised.

One method is using https://github.com/leolin310148/ShortcutBadger/ to off-load setting the badges, but it seems that the most popular phones use `android.intent.action.BADGE_COUNT_UPDATE` to set the badge. Example from [DefaultBadger](https://github.com/leolin310148/ShortcutBadger/blob/master/ShortcutBadger/src/main/java/me/leolin/shortcutbadger/impl/DefaultBadger.java):

```java
private static final String INTENT_ACTION = "android.intent.action.BADGE_COUNT_UPDATE";
private static final String INTENT_EXTRA_BADGE_COUNT = "badge_count";
private static final String INTENT_EXTRA_PACKAGENAME = "badge_count_package_name";
private static final String INTENT_EXTRA_ACTIVITY_NAME = "badge_count_class_name";

public void executeBadge(Context context, ComponentName componentName, int badgeCount) throws ShortcutBadgeException {
   Intent intent = new Intent(INTENT_ACTION);
   intent.putExtra(INTENT_EXTRA_BADGE_COUNT, badgeCount);
   intent.putExtra(INTENT_EXTRA_PACKAGENAME, componentName.getPackageName());
   intent.putExtra(INTENT_EXTRA_ACTIVITY_NAME, componentName.getClassName());
   context.sendBroadcast(intent);
}
```

Would it be possible to implement `push.setApplicationIconBadgeNumber()`, using this intent?

barryvdh:
Adds https://github.com/leolin310148/ShortcutBadger support for Android badges, using `setApplicationIconBadgeNumber()` Fixes #622

Update: Now also handles background notifications

- [x] Use ShortcutBadger to update badges
- [x] Implement `setApplicationIconBadgeNumber()` call directly, similar to iOS (using different thread)
- [x] Update badge from GCM `count` property

### DIFF
--- a/src/android/com/adobe/phonegap/push/PushPlugin.java
+++ b/src/android/com/adobe/phonegap/push/PushPlugin.java
@@ -23,6 +23,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 
+import me.leolin.shortcutbadger.ShortcutBadger;
+
 public class PushPlugin extends CordovaPlugin implements PushConstants {
 
     public static final String LOG_TAG = "PushPlugin";
@@ -179,6 +181,18 @@ public class PushPlugin extends CordovaPlugin implements PushConstants {
                     }
                 }
             });
+        } else if (SET_APPLICATION_ICON_BADGE_NUMBER.equals(action)) {
+            cordova.getThreadPool().execute(new Runnable() {
+                public void run() {
+                    Log.v(LOG_TAG, "setApplicationIconBadgeNumber: data=" + data.toString());
+                    try {
+                        setApplicationIconBadgeNumber(getApplicationContext(), data.getJSONObject(0).getInt(BADGE));
+                    } catch (JSONException e) {
+                        callbackContext.error(e.getMessage());
+                    }
+                    callbackContext.success();
+                }
+            });
         } else {
             Log.e(LOG_TAG, "Invalid action : " + action);
             callbackContext.sendPluginResult(new PluginResult(PluginResult.Status.INVALID_ACTION));
@@ -216,6 +230,14 @@ public class PushPlugin extends CordovaPlugin implements PushConstants {
                 Log.v(LOG_TAG, "sendExtras: caching extras to send at a later time.");
                 gCachedExtras = extras;
             }
+        }
+    }
+
+    public static void setApplicationIconBadgeNumber(Context context, int badgeCount) {
+        if (badgeCount > 0) {
+            ShortcutBadger.applyCount(context, badgeCount);
+        } else {
+            ShortcutBadger.removeCount(context);
         }
     }
 
--- a/plugin.xml
+++ b/plugin.xml
@@ -86,6 +86,7 @@
 
         <framework src="com.android.support:support-v13:23+" />
         <framework src="com.google.android.gms:play-services-gcm:+" />
+        <framework src="me.leolin:ShortcutBadger:1.1.4@aar" />
 
 		<source-file src="src/android/com/adobe/phonegap/push/GCMIntentService.java" target-dir="src/com/adobe/phonegap/push/" />
 		<source-file src="src/android/com/adobe/phonegap/push/PushConstants.java" target-dir="src/com/adobe/phonegap/push/" />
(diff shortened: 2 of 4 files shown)


barryvdh:
Ideally we could add a `badge` option to the Android init, and then set the badge when possible, so we avoid any BC breaks.

src/android/com/adobe/phonegap/push/PushPlugin.java
```diff
@@ -179,6 +181,20 @@ public void run() {
                     }
                 }
             });
+        } else if (SET_APPLICATION_ICON_BADGE_NUMBER.equals(action)) {
```
yossale: If I understand correctly, this code only runs when the app is opened, not when the push message arrives (That's what happens on my Device, anyway). If this is indeed the case, it can't update the counter badge until the app opened, which doesn't solve the bug - no?

barryvdh: Not fully no, that's what I said, not really sure how. But with content available, you could still update in background.
But for background, this could work similarly right? Not sure how to fix it otherwise.

yossale: I've solved it in this branch: 
https://github.com/silo-co/phonegap-plugin-push
You can read about it here:
http://www.yossale.com/Ionic-android-icon-badges/

barryvdh: Okay so your patch (https://github.com/phonegap/phonegap-plugin-push/compare/master...silo-co:master) handles the updating of the push in the background, and mine in foreground for `setApplicationIconBadgeNumber` right?

So we can combine both methods, right? Can you submit a PR to my branch to add that?

yossale: I don't think we need to merge them - the code I wrote handles both cases, because it's called anyway. I'm not sure that handling the same issue from different places in the code is wise

barryvdh: No the first method happened only when you called `setApplicationIconBadgeNumber()` directly, right?
I've abstracted it to a public static function in the PushPlugin class, and added it to be called when a `count` it available. Similar to your branch, but when it's `null`, it doesn't change.
Also, you can call it directly (eg. without push, but with `setApplicationIconBadgeNumber()` directly)

barryvdh:
This PR is updated to support both GCM intents + direct setting with `setApplicationIconBadgeNumber()`. When the `count` property is set in a GCM message, it should set the badge directly now.

barryvdh:
I have now used a different thread for handling `setApplicationIconBadgeNumber()` inside the app, similar to other commands.

macdonst:
@barryvdh can you please retest by using the 1.7.x branch of the plugin? I don't actually have any phones that support badging. I've tested on a OnePlusOne and Nexus 5 with no ill affects.

barryvdh:
Setting the badge works on my HTC M8. It also works with push messages, with one change:
Sending background badge updates (without message/title) doesn't work anymore, because you set the badge number together with creating the notification. So without message/title, the badge will not be updated (not even with content-available).
I don't know if this intentional?

macdonst:
@barryvdh ah, good point. I'll make a change and let you know when to re-test.

macdonst:
@barryvdh okay, I made the change in the v1.7.x please retest.

macdonst:
@barryvdh have you got a chance to re-test the v1.7.x branch?

barryvdh:
Not fully yet, but the way you set the icon didn't really change much since last time I tested..

macdonst:
@barryvdh I thought I made the change so that the badge would be set on every incoming message.

barryvdh:
Yeah I meant that you didn't change the way you actually called the badger plugin, just the logic for parsing the message etc. So even if you didn't test it actually sets the badge, you can verify the logic, right?
But I'll check soon.

YaroslavG:
hi guys,
can I see budge icon on my android version 4.2.2 ?

macdonst:
@barryvdh merged into master now, thanks for the code.

barryvdh:
Thanks!

GurubakshSingh:
Thanks a lot i am new on cordova and working on asus phone and its working. After spending 4 days. Your great

Issue #481, opened 2016-01-07
Error Permission name C2D_MESSAGE is not unique on release android


Alexintosh:
I'm getting this error when I build for release android.

If I remove the plugin everything is working fine, any ideas?

Cordova version 5.4.0

![screen shot 2016-01-07 at 15 21 45](https://cloud.githubusercontent.com/assets/776519/12174083/67bb0b60-b552-11e5-8459-4e2276ad1f49.png)

macdonst:
Did you uninstall the debug version of your app first?

Alexintosh:
Hi @macdonst 
What do you mean for unistall? 
I'm just building with cordova build --release android

macdonst:
@Alexintosh I can't reproduce this. What version of the plugin are you using?

Alexintosh:
I'm using version 1.5.1
I'm trying to understand what's going on, I'll keep you updated.

macdonst:
@Alexintosh I'd try clearing out the platforms and plugins directory and starting clean.

Zeychiel:
Hi,
I encountered the same error. Clearing out platforms and plugins did'nt work, the only workaround is...to disable phonegap-plugin-push.
My environment is brand new, sdk, plugin and cordova are last versions.

VinceJou:
Hi,
I encountered the same error. Clearing out platforms and plugins did'nt work, the only workaround is...to disable phonegap-plugin-push.
My environment is brand new, sdk, plugin and cordova are last versions.

macdonst:
@VinceJou do you use any other plugins that may be adding C2D_MESSAGE?

Alexintosh:
For some reason I had some hardcoded permissions in the Manifest, once cleaning those it worked.

VinceJou:
@macdonst Yes, I have another application using it

FerryDewe:
My solution, is to remove the duplicate plugin.
In my case it was, between this com.phonegap.plugins.PushPlugin or phonegap-plugin-push
Run this before the release, will remove the double permission in the Manifest
> cordova plugin rm com.phonegap.plugins.PushPlugin

macdonst:
@VinceJou right, you can only have that in the AndroidManifest.xml once. What is the other plugin you are using?

VinceJou:
@macdonst Hi, sorry for the delay...the other application using it is ms-azure-mobile-engagement.
[...]

PR #625, opened 2016-02-16
Use standard cordova PACKAGE_NAME variable in place of applicationId

cfsnyder:
Standard variable for this is documented here: https://cordova.apache.org/docs/en/4.0.0/plugin_ref/spec.html#$package_name


macdonst:
@cfsnyder this works and it's a valid way to fix the issue but I'm wondering if we need to merge it as what's there isn't broken and in the new Cordova-Android they add the same fix so we won't need the push.gradle file soon.

https://github.com/apache/cordova-android/commit/fb9cf60c4100d5bc6f9320747611da1ca0ced289

So what benefit does merging this PR give us? I'm not being sarcastic, I'm asking seriously.

cfsnyder:
@macdonst the current push.gradle solution wasn't working in our build process, for whatever reason. I came a across the standard $PACKAGE_NAME variable, found that it fixed my issue and was simpler than the current approach, so I thought I'd submit a PR to suggest the change. If others aren't having issues with the push.gradle solution and Cordova-Android is adding the same fix, I agree that it isn't necessary for you to merge this.

macdonst:
@cfsnyder how are you building? Gradle is the future of all the Android stuff.

cfsnyder:
@macdonst we're using the standard gulp-cordova plugins to initiate the build with a gulp task, which means we are using the gradle build system implicitly. I'm not sure if push.gradle was never being run or it may be possible that its output was being overwritten by another stage of the build process. I would have to dive into it further to be sure.

macdonst:
@cfsnyder I'm not going to merge this but I do appreciate the pull request. Can you send me more info on your build process as it may be something I can help fix in Gulp Cordova.

macdonst:
@cfsnyder well as it turns out this change looks like it will fix #481 so I will be merging.

cfsnyder:
@macdonst great glad it could be used.

macdonst:
@cfsnyder yeah, it works around a bug in plugman so why not :)

Thanks again for the submission!

PR #645, opened 2016-02-24
Add coldstart property for iOS

kenvunz:
Trying to mimic the same `coldstart` behaviour of the android version.

Like in the android version, `coldstart` has 3 different states instead of just `true | false`

- `undefined` (`coldstart` property doesn't exist in the `addtionalData` object) - `on('notification')` is triggered when the app is in either **background** or **foreground** mode

- `true` - `on('notification')` is triggered when the app is launched from a notification, and the app is in **suspended** mode

- `false` - `on('notification')` is triggered when the app is launched from a notification, and the app is in **background** mode

I've tested it fairly thoroughly with iOS 9.2 and it seems to be working fine. I'm not an Objective-C expert so I might have miss something, feel free to point them out and I'll try to fix them.

### DIFF
--- a/src/ios/AppDelegate+notification.m
+++ b/src/ios/AppDelegate+notification.m
@@ -44,11 +44,15 @@ static char launchNotificationKey;
 // to process notifications in cold-start situations
 - (void)createNotificationChecker:(NSNotification *)notification
 {
+    NSLog(@"createNotificationChecker");
+
     if (notification)
     {
         NSDictionary *launchOptions = [notification userInfo];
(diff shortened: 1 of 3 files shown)


macdonst:
@kenvunz thanks for the PR but I already implemented coldstart in the v1.6.x branch. Please take a look over commit https://github.com/phonegap/phonegap-plugin-push/commit/ef8759f2a3c3c8117aab9ca71086b389ed2dde4d and let give me some comments.

Issue #447, opened 2015-12-19
notification_applicationDidBecomeActive method


code-smith:
In AppDelegate+notification class the method notification_applicationDidBecomeActive replaces applicationDidBecomeActive but its not getting called when app becomes active.

Earlier i used applicationDidBecomeActive to reset badges when opened directly from clicking the app icon, and had few other actions but it doesnt look like i can do that here?
is this a bug or by design , Can anyone please help me here? 

Thanks,
Sid

macdonst:
I'm going to revert the code that caused this problem. Hold tight.

code-smith:
Thanks..

macdonst:
@code-smith published 1.5.2 to NPM. It should show up shortly. In the meantime you can install from the giturl.

Issue #427, opened 2015-12-15
[ios] "the right way" to swizzle


athibaud:
it seems using `method_exchangeImplementations()` can have some [undesired implications](https://blog.newrelic.com/2014/04/16/right-way-to-swizzle/) and the *right* way to do method swizzling is to use `method_setImplementation()`

shouldn't run into the implications described in that article seeing the use of this plugin is more of a *standalone application* use rather than *library* use but still.
i could implement and pr this if needed.

cheers

macdonst:
PR merged

athibaud:
i imagine you mean this pr https://github.com/phonegap/phonegap-plugin-push/pull/426 (thanks for merging :-)) but the code still uses `method_exchangeImplementations()` rather than `method_setImplementation()`

the pr i'm referring too in **this** issue does not yet exist :-)

macdonst:
@athibaud my bad

PR #661, opened 2016-03-01
[ios] better swizzling

devgeeks:
This still isn't the "right way" linked in #427, but seems to be safe
and void of side effects due to the namespacing of all the methods.

It has been tested to work side by side with another plugin that also
swizzles AppDelegate's applicationDidBecomeActive.

1. wrapped swizzling in dispatch_once
2. first attempts class_addMethod / class_replaceMethod but if that
fails falls back to method_exchangeImplementations.
3. swizzled_init => pushPluginSwizzledInit (if other plugins attempt to
swizzle init and use the same name, there seems to be a collision)
4. added an observer for UIApplicationDidBecomeActiveNotification
instead of overriding applicationDidBecomeActive
5. applicationDidBecomeActive => pushPluginOnApplicationDidBecomeActive
(and it becomes a notification observer instead)

See http://nshipster.com/method-swizzling/ and "Avoid collisions" under
"Considerations".

This will need testng by someone that understands the plugin better than
I do, but it should not be changing the core functionality at all. The
pushPluginOnApplicationDidBecomeActive method *is* called when the
application becomes active, etc.

Goes a long way toward fixing #427 without regressing #447, etc.

### DIFF
--- a/src/ios/AppDelegate+notification.h
+++ b/src/ios/AppDelegate+notification.h
@@ -12,7 +12,7 @@
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken;
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error;
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:( void (^)(UIBackgroundFetchResult))completionHandler;
-- (void)applicationDidBecomeActive:(UIApplication *)application;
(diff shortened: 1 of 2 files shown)


macdonst:
@devgeeks thanks, I'll be testing it out later today. This'll become part of the 1.6 release I'm busy on right now so your users will soon have a fix for this.

devgeeks:
Actually... funnily enough, only one of the plugins needed to implement this, so even with push 1.5.3, if my plugin has this patch, they work together. But it would be best if the push plugin didn't step on say local notifications etc...

macdonst:
@devgeeks works like a charm. I merged it into the v1.6.x branch and will close the PR when I merge that into master.

devgeeks:
:+1:

PR #690, opened 2016-03-10
docs: Explain `forceShow` in more detail

jakub-g:
This is just docs change. I did not realize the full consequences of `forceShow` and I think it can be useful to the others as well


macdonst:
@jakub-g LGTM, thanks!

jakub-g:
Cool, thanks!

Issue #158, opened 2015-09-21
"Notification" Event Not Firing


ryanpager:
So this is the use case im following:

a. Push notification goes to phone
b. User clicks on notification from lock, which opens the app & triggers the 'notification' event
c. Do some obligatory action based on additionalData

This is working perfectly, except it seems like the 'notification' event does not get triggered ever again if the user closes the application through the app launcher (square button & swipe).

So a run down of the problem in short:

a. User opens application, push is registered. 
b. 'Notification' events now get triggered as expected.
c. User closes the application through the app launcher.
d. 'Notification' events no longer get triggered....ever.

Any ideas on this one?

fredgalvao:
Is this on Android?

ryanpager:
Yea, I'm testing this on Android 5.1.1 on a Nexus 5.

I also tried reregistering the device on cold start which fixes the future
notification events (great) but not existing ones already issued.

On Monday, September 21, 2015, Frederico Galvão <notifications@github.com>
wrote:

> Is this on Android?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phonegap/phonegap-plugin-push/issues/158#issuecomment-142124419>
> .
>


-- 
*Ryan Page*
*"In Soviet Russia, Branch Merge You!"*
*w | *http://sudobangbang.io* e |* ryanpager@gmail.com *p |* 469-404-7483

macdonst:
@ryanpager after you force close the app do you still see the new push messages in the notification shade?

ryanpager:
@macdonst Oddly enough yea -- push notifications still work exactly as expected, they just dont trigger the 'notification' event regardless of if the app is in foreground or background.

macdonst:
@ryanpager Okay, that's actually the behavior I would expect. When you close the app through the app launcher you are effectively killing your app so the on('notification') handler has been removed. I'm going to look at waking up the app on incoming notifications in the future but for now this is not handled. 

PR's on this topic are welcome.

app2naklabs:
@ryanpager I am also having similar issue and using phonegap build so can you tell me how you able to receive notification even after force close.

ryanpager:
[...]

PR #703, opened 2016-03-14
Support android shortcut badges on supported devices when app is closed

yossale:
Update android shortcut badge even when app is closed (on supported devices - HTC, LG, Samsung)

## Description
Adding android shortcut badges support. Similarly to the [cordova-plugin-badge](https://github.com/katzer/cordova-plugin-badge) behaviour, only it works even when the app is closed, and updated from the Java, not the js. 

## Related Issue
["Notification" Event Not Firing When Closed Through App Launcher](https://github.com/phonegap/phonegap-plugin-push/issues/158)
[Android badge count](https://github.com/phonegap/phonegap-plugin-push/issues/622)

## Motivation and Context
Most major Android Launchers now support badges, which are a useful way to keep users engaged with one's app, specially when users are not visiting it too often. However, in the current situation, badges are only updated when the app is either in the foreground/background, which is usually not the case (Specially if your app is a "twice a week kind" of app)

## How Has This Been Tested?
It has been tested on Samsung galaxy S5, HTC X9, and One Plus (which doesnt support badges, we tested to make sure nothing breaks in that case)

## Types of changes
- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -23,7 +23,7 @@
 	<engines>
         <engine name="cordova" version=">=3.6.3" />
         <engine name="cordova-android" version=">=4.0.0" />
-        <engine name="cordova-ios" version=">=4.0.0" />
+        <engine name="cordova-ios" version=">=3.9.2" />
     </engines>
 
     <preference name="SENDER_ID" />
--- a/plugin.xml
+++ b/plugin.xml
@@ -86,6 +86,7 @@
 
         <framework src="com.android.support:support-v13:23+" />
         <framework src="com.google.android.gms:play-services-gcm:+" />
+        <framework src="build.gradle" custom="true" type="gradleReference" />
 
 		<source-file src="src/android/com/adobe/phonegap/push/GCMIntentService.java" target-dir="src/com/adobe/phonegap/push/" />
 		<source-file src="src/android/com/adobe/phonegap/push/PushConstants.java" target-dir="src/com/adobe/phonegap/push/" />
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -59,7 +59,6 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
     @Override
     public void onMessageReceived(String from, Bundle extras) {
         Log.d(LOG_TAG, "onMessage - from: " + from);
-        ShortcutBadger.applyCount(getApplicationContext(), 3);
         if (extras != null) {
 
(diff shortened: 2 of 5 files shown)


plugin.xml
```diff
@@ -23,7 +23,7 @@
 	<engines>
         <engine name="cordova" version=">=3.6.3" />
         <engine name="cordova-android" version=">=4.0.0" />
-        <engine name="cordova-ios" version=">=4.0.0" />
+        <engine name="cordova-ios" version=">=3.9.2" />
```
macdonst: Nope, for version 1.6.0 and greater of this plugin we'll only support cordova-ios 4.0.0 or greater. Please revert line.

plugin.xml
```diff
@@ -86,6 +86,7 @@
 
         <framework src="com.android.support:support-v13:23+" />
         <framework src="com.google.android.gms:play-services-gcm:+" />
+        <framework src="build.gradle" custom="true" type="gradleReference" />
```
macdonst: Instead of including a custom build.gradle file why not use a framework tag:

`<framework src="me.leolin:ShortcutBadger:1.1.4@aar" />`

src/android/com/adobe/phonegap/push/PushPlugin.java
```diff
@@ -4,8 +4,8 @@
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.util.Log;
 
+import android.util.Log;
```
macdonst: Remove unused import.

macdonst:
@yossale looks good but please make the required changes and squash your commits.

barryvdh:
PR #624 offers similar functionality, but also supports updating the badge from the javascript app itself.

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -191,16 +192,32 @@ private Bundle normalizeExtras(Bundle extras) {
         return newExtras;
     }
 
+    private void updateBadge(Context context, String badge) {
+
+        int count = badge == null ? 0 : Integer.parseInt(badge);
```
barryvdh: This would make it impossible to keep the badge count unchanged (eg. when it's not set)

ejazdragonflycreation:
after install remove [Phone-gap](https://github.com/phonegap/phonegap-plugin-push/) and add [Silo-co](https://github.com/silo-co/phonegap-plugin-push/) , when push notification comming. the runtime error generates and app close on android

macdonst:
@yossale not going to merge as it is superceeded by #624 but the functionality will be coming in 1.7.0.

Issue #689, opened 2016-03-10
Remove sender id from PushNotification init Android options


restarian:
#### The sender id in version 1.6.0+ is now available as a plugin variable, thusly should be omitted from app front-end script.

```
[Notification_Peripheral] Plugin error with the PushNotification process {
  "0": Error: No value for senderID
    at fail (file:///android_asset/www/plugins/phonegap-plugin-push/www/push.js:56:49)
    at Object.cordova.callbackFromNative (file:///android_asset/www/phonegap.js:293:52)
    at processMessage (file:///android_asset/www/phonegap.js:1079:17)
    at processMessages (file:///android_asset/www/phonegap.js:1102:9)
    at pollOnce (file:///android_asset/www/phonegap.js:971:9)
    at pollOnceFromOnlineEvent (file:///android_asset/www/phonegap.js:958:5)
}
```

macdonst:
@surgemcgee Yes, that is the plan but I'm waiting until a 2.0.0 release as I don't want to make a breaking API change.

jakub-g:
In fact just FYI it was already a breaking change with 1.6.0 release kind of.

I'm using ionic and a build script does `ionic state restore` to install platforms and plugins (which reads `package.json` and install stuff with proper versions and variables)

When I bumped version from 1.5.3 to 1.6.0 in `package.json` and let my scripts run, it failed due to missing SENDER_ID configuraton.

I had to change

```
 "phonegap-plugin-push@1.5.3",
```
to
```
{
  "variables": {
    "SENDER_ID": "........"
  },
  "locator": "phonegap-plugin-push@1.6.0",
  "id": "phonegap-plugin-push"
},
```

If someone had `^1.5.3` in `package.json`, or unspecified version, their automated CI build would take `1.6.0` and fail.

That's why I prefer strict versions in `package.json` to have control over the update process :)

ujjal008:
Ionic Core: init
ionic.io.bundle.min.js:2 Ionic Core: searching for cordova.js
ionic.io.bundle.min.js:2 Ionic Core: attempting to mock plugins
ionic.io.bundle.min.js:2 Ionic Push: register

But when I install on device not get devicetoken. call back function not working:

.run(function($ionicPlatform) {
  $ionicPlatform.ready(function() {
    alert("push")
    var push = new Ionic.Push({
      "debug": true
    });

    push.register(function(token) {
      alert(token);

      //alert("Device token:",token.token);
[...]

Issue #693, opened 2016-03-11
Unable to install plugin on both iOS/Android


dilipsarkar:
I was trying to verify the fix for issue` #688` but I'm unable to installed the plugin. I get the below error on both iOS and Android.

```
cordova plugin add CustomPlugins\phonegap-plugin-push-master
Error: Variable(s) missing (use: --variable SENDER_ID=value).

```
I'm using `cordova 5.4.1`. Please help.

jakub-g:
See https://github.com/phonegap/phonegap-plugin-push/issues/689#issuecomment-195316034

when you install from command line, pass `--variable SENDER_ID=1234` as required. If you use Android and GCM, you already know sender id (and you already pass it in `android` config when registering the plugin). Otherwise put a fake value.

dilipsarkar:
Thanks for your quick response. 

Just a clarification, does this mean we don't have to specify `senderID `for android while invoking `PushNotification.init()` API ?

```
var push = PushNotification.init({
    android: {
      senderID: "12345679
    },
    ios: {
        alert: "true",
        badge: "true",
        sound: "true"
    },
    windows: {}
});
```

jakub-g:
in 1.6 you have to do it in both places, in 2.0 only in `package.json`

dilipsarkar:
Great, thanks

PR #714, opened 2016-03-16
Feature/default android senderid

cmprescott:
## Description

Store Android senderID set at installation in the AndroidManifest.xml. Use value in AndroidManifest.xml as the default.

## Related Issue

Fixes Issues #689 & #693 without breaking existing usages.

## Motivation and Context

Allows other plugins to use this as dependency to automagically get the pushId onDeviceReady

In the test app I ran this code

```javascript
    onDeviceReady: function() {
        app.receivedEvent('deviceready');

            android: {}
        });

        push.on('registration', function(data) {
            console.log('registrationId: ' + data.registrationId);
        });

        push.on('notification', function(data) {
            console.log('msg: ' + data.message);
            console.log('title: ' + data.title);
        });

        push.on('error', function(e) {
            console.log('error: ' + e.message);
        });
    },
```

image

## Types of changes

## Checklist:

I could document this new feature, but might be confusing until you decide to deprecate init with senderID in the 2.0 release. So I choose not to.


marianoarga:
The idea is to store the "senderId" into AndroidManifest.xml?

cmprescott:
It is. I unsuccessfully tried several other ways to store it. Getting from strings.xml runs into problems with each app has a different namespace & that would require some messy plugin hooks. I tried to figure out CordovaPreferences, but those don't seem documented for this kind of use. I also thought about switching to using the 'google-services.json' file, but that seemed like too big a change to incorporate into a `.` release.

marianoarga:
And how to setup de senderId inside AndroidManifest.xml? I think you missed that in the description maybe.

cmprescott:
It's already documented on https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/INSTALLATION.md . You've had to pass this variable in since 9e3b843c2dbad1cc2546aede4fb0b33dd7588451 . This just makes it so you don't have to pass in at installation time AND have to init the javascript with it. See issue #689.

marianoarga:
I see, I was using it from init. Thank you.

cmprescott:
@macdonst Any feedback on if we can incorporate this? I can split into 2 pull requests if you prefer 1:1 issue to pull request. I combined because they are closely tied together.

PR #773, opened 2016-04-01
Update INSTALLATION.md

theaccordance:
Update of installation instructions to account for installing plugins via the `phonegap prepare` method.

## Motivation and Context
Adding this plugin to the config.xml file using the `--save` argument omits the variable.  The CLI silently skip installing the plugin using `phonegap prepare` unless the SENDER_ID variable is also defined in the config.xml file.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -50,6 +50,14 @@ Where the `XXXXXXX` in `SENDER_ID="XXXXXXX"` maps to the project number in the G
   ]
 ```
 
+> Note: You need to specify the SENDER_ID variable in your config.xml if you plan on installing/restoring plugins using the prepare method.  The prepare method will skip installing the plugin otherwise.
+
+```
+<plugin name="phonegap-plugin-push" spec="1.6.0">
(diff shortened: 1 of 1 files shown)


macdonst:
@theaccordance what version of phonegap are you using? When I do:

`phonegap plugin add phonegap-plugin-push --variable SENDER_ID=XXXXXXXX`

the variable is saved into config.xml correctly.

theaccordance:
I stand corrected (at least with the current version), the `--save` argument does add the tag properly.  So no pragmatic bug to resolve with this.  I've been working on a complete overhaul of the cordova dependencies on our project, it's possible I copied the tag incorrectly to the template we use to compile the config.xml file before kicking off the build.  

I think it might still be valuable to have some type of language in the documentation to clarify how the plugin needs to be defined in the config.xml file. This is the first plugin in our project that has any type of the child tag in the plugin manifest, wasn't something intuitive to debug when we were trying to figure out why it wasn't loading properly.

macdonst:
@theaccordance will do.

PR #864, opened 2016-05-04
Adding support for background notifications on Windows

rakatyal:


nikhilkh:
👍

docs/PAYLOAD.md
```diff
@@ -924,3 +925,25 @@ The default handling can be disabled by setting the 'cancel' property in the not
...
+    <visual>
+        <binding template="ToastImageAndText01">
+            <image id="1" src="ms-appx:///images/redWide.png" alt="red graphic"/>
+            <text id="1">Hello World!</text>
+        </binding>
+    </visual>
+</toast>
+```
+
+This launch attribute string is passed on to the app as data.launchArgs through the on('notification') handler. It's important to note that due to the Windows platform design, the other visual payload is not available to the handler on cold start. So notification attributes like message, title etc. which are available through the on('notification') handler when the app is running, won't be available for background notifications.
```
fredgalvao: Might as well add `data.launchArgs` to the `push.on('notification', callback)` callback parameters on the docs too, to keep it consistent. Updating our typescript definition on DefinitelyTyped would also be welcome, but that can wait I guess.

rakatyal: Yup. I'll do that.

Issue #897, opened 2016-05-18
'notification' event not triggering in background notification Android


Moghul:
### Expected Behaviour
1) Get a notification
2) Tap notification
3) App opens, 'notification' event triggers.
### Actual Behaviour
1) Get a notification
2) Tap notification
3) App opens, 'notification' event does not trigger.
#### Steps to Reproduce 
Send the payload below
Follow steps mentioned above
#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Android 6.0.1
#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
Cyanogen Mod
Oneplus One
#### Cordova CLI version and cordova platform version
Cordova CLI 5.4.1
#### Plugin version
Plugin Version 1.6.3
#### Sample Push Data Payload
```
{   "priority" : "high",
    "notification": {
        "title": "Test Notification",
        "body": "This offer expires at 11:30 or whatever",
        "notId": 10
    },
    "data" : {
        "surveyID": "ewtawgreg-gragrag-rgarhthgbad",
        "content-available": "1"
    },
    "to": "id here"
}
```
#### Sample Code that illustrates the problem
                notification = PushNotification.init({
                    "android": {"senderID": "id", "sound": "true", "vibrate": "true"},
                    "ios": {"senderID": "id", "gcmSandbox": "true", "alert": "true", "badge": "true", "sound": "true"},
                    "windows": {}});
                notification.on('registration', onRegistration);
                notification.on('notification', onNotification);
                notification.on('error', onError);

 var onNotification = function (data) {
            console.log("NOTIFICATION IS " + JSON.stringify(data));
}
#### Logs taken while reproducing problem
05-18 11:04:34.014 2691-2691/? D/StatusBar: Clicked on content of 0|APP-ID-HERE|0|GCM-Notification:267160844|10194

Note that this behavior ONLY happens on Android when the 'notification' block is present in the sample push.
Only including "data" makes it work fine.
Removing 'notification' from the json is not an option because without it, it doesn't work on iOS.
With the previously mentioned payload, it works perfectly fine on iOS.

Things I've tried: Moving around the "content-available": "1" line as well as changing it to "content_available" : true and "true".

Possible solution considered: Somehow checking if the user is from iOS and having different payloads but I haven't looked into whether GCM offers that functionality.

EDIT:
Alright, looking at the docs, this is totally something that should work on Android:

Hybrid messages with both notification and data payload

[...]

Issue #883, opened 2016-05-13
Intel XDK, Gradle


Moghul:
I'm building my app using Intel XDK and they can't build if the plugin has a gradle file in it.
From what I can see, the gradle edits what is basically 2 lines in the manifest. It adds the app id to the manifest on those lines.

Is it enough for me to remove the gradle file, remove the reference, and edit those lines in the plugin's xml file with my app id? Will the plugin work?

EDIT: should the app ID be backwards?

Well, I just tried it and it doesn't work, the log says

Package some.example.app attempting to redeclare permission some.example.com.permission.C2D_MESSAGE already owned by com.intel.appx.example.xwalk16

EDIT 2: Fixed mistyped app id. Android built fine but iOS won't build. All it says is failed to install push plugin.

EDIT 3: Found #817, looking into that. Seems like I answered a bunch of my own questions.

macdonst:
@Moghul yeah, the reason I had to switch back to using the gradle file is you get those permission errors if two plugins both want to add the same permission. The only thing to do at this point is to ask IntelXDK team to support gradle files in plugins. Feel free to send them my way.

Moghul:
@macdonst But if I don't have 2 plugins asking for the same permissions it's fine if I just modify the plugin and keep it in mind for the future, correct?

ppetree:
I hate to bring this zombie to the top but what was the ultimate outcome? I'm getting ready to start down this path and want to make sure it works!

P.

macdonst:
@ppetree check out this issue where @tony-- goes through the whole thing. https://github.com/phonegap/phonegap-plugin-push/issues/926#issuecomment-223191041

Moghul:
If you need to build with the Intel XDK, for now, you'll have to remove the gradle file and refactor the ${application} in the plugin.xml with your app's app ID. Note that you'll need to use the debug module's app ID (something like com.intel.appx.[companyname].xwalk16). It's a pain in the ass but it works.

macdonst:
Yeah, I can't add Gradle support to IntelXDK but @tony-- did a great write up on how to compile this plugin with the XDK.

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/PHONEGAP_BUILD.md#intelxdk-support

PR #904, opened 2016-05-20
Possible fix for hybrid notification payloads on Android with possible solution for deleting notifications.

Moghul:
Added a class that handles background notifications for notifications that come in "notification" payload instead of "data" payload.
## Description

Added a new class that extends GcmReceiver. The code for it is a mirror copy of the GCMIntentService class with the minor difference of adding a context and using it since it's not an Action.

Also added a function to the same class that can delete a notification if the value "delete" : notificationID is present in the "data" payload on Android.
## Related Issue

#897 

## Motivation and Context

Solves 2 issues: Notification deletion on Android and background notifications on Android.
## How Has This Been Tested?

Pushed notifications for the app while it was
1) Closed completely
2) Closed but present in the recent apps screen
3) Open and in the foreground

Pushed notifications with delete keys where the value was the notId of a still available notification.

## Types of changes

## Checklist:

### DIFF
--- a/BackgroundNotificationHandler.java
+++ b/BackgroundNotificationHandler.java
@@ -1,0 +1,653 @@
+package com.adobe.phonegap.push;
+
+import com.google.android.gms.gcm.GcmReceiver;
+import android.content.SharedPreferences;
+import android.content.SharedPreferences.Editor;
+import android.preference.PreferenceManager;
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.ContentResolver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.content.res.AssetManager;
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Color;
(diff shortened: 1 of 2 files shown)


fredgalvao:
Why did you delete `<framework src="push.gradle" custom="true" type="gradleReference" />` from `plugin.xml`?

Moghul:
@fredgalvao That was a slight oversight, sorry. The build environment I am using can't handle Gradle so I removed the file and its references and replaced ${application} with my own app ID. I changed the latter back for the push but I forgot about the reference...
See #883 
Of course, feel free to re-add it.

fredgalvao:
@Moghul I see. I think we can leave solving the build process on XDK to issue #883, but that's just my opinion.

Regardless, @macdonst 's words will probably be the final ones.

Moghul:
@fredgalvao Of course. As for #883, the "fix" wasn't supposed to be in here. Preferably, the guys at Intel would solve it themselves. This was just something I needed for myself.

macdonst:
@Moghul can you rebase this PR with master and fix the plugin.xml issue and re-submit? Shouldn't need to touch every line in plugin.xml.

Moghul:
See #941

Issue #929, opened 2016-05-25
Migrate to Firebase Cloud Messaging


macdonst:
Google is moving from GCM to FCM and this plugin should follow suit. 

https://developers.google.com/cloud-messaging/android/android-migrate-fcm

I may end up supporting a 1.x and 2.x stream where the 2.x stream is FCM as it will be awhile before everyone can update to using FCM.

skyprimer:
Great.! we are following the topic...

getqd:
Same. Looking forward to integrating FCM into our apps!

Maximvdw:
+1

lorenzob-incomedia:
+1

raphaguasta:
Great!! Thanks for this!

JeremyColton:
FCM makes iOS APNS obsolete since FCM is reliable and free?

sn0w:
+1

Any ETA on this?

zishon89us:
+1

waiting for it , http://stackoverflow.com/questions/37653251/firebase-cloud-messaging-with-ionic-app/

lukebpowell:
+1

srikantk:
+1

vasanthns:
+1 following..

lurainsin:
+1
I'm looking forward to use FCM in my new app!

kumpanen:
+1

MaximKipot:
+1. 
really excited about that!

JeremyColton:
Please stop adding "+1" - you are spamming everyone!  Just click the "subscribe" button on the right-hand-side!

GitFr33:
I am just getting started with push notifications. Would you advise going ahead and setting up GCM and migrating later, or waiting until FCM is supported?

GrzegorzWarzecha:
+100!

ktzanettis:
Anyone know when this issue will be committed?

jcesarmobile:
When? once it's ready

macdonst:
@ktzanettis you can always help out by testing the v2.0.x branch or submitting a PR.

glasset:
I have tested on v2.0.x(07fa81) currently works fine (tested only fcm feature), if something bad happen i will report

stephan-nordnes-eriksen:
@glasset Great to hear! I would like to try it out and help if possible. Is the setup any different than gcm?

glasset:
@stephan-nordnes-eriksen yes same as gcm setup.

[...]

PR #975, opened 2016-06-09
Added partial Android FCM support

jcesarmobile:
Removed permissions and classes unneeded
Made required changes to work with FCM

## Related Issue
Not linking the issue as it's not fully completed and only the android part

## Motivation and Context
Because FCM is the new thing

## How Has This Been Tested?
Tested on android 5 and android 6 devices

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

Known issue. The first time you launch the app you won't get the push token on the javascript part, needs changes to pick it when onTokenRefresh is called


macdonst:
@jcesarmobile it looks pretty good to me, I'll leave a few comments on specific lines. Can you rebase with the v2.0.x branch and resubmit the PR as it won't merge clean?

plugin.xml
```diff
@@ -40,60 +40,39 @@
         <config-file target="res/values/strings.xml" parent="/resources">
             <string name="google_app_id">$SENDER_ID</string>
```
macdonst: @jcesarmobile do we need this anymore? Is it just contained in the google-services.json file?

plugin.xml
```diff
@@ -40,60 +40,39 @@
...
-            <service
-                android:name="com.adobe.phonegap.push.RegistrationIntentService"
-                android:exported="false">
-            </service>
 		</config-file>
 
         <framework src="push.gradle" custom="true" type="gradleReference" />
         <framework src="com.android.support:support-v13:23+" />
-        <framework src="com.google.android.gms:play-services-gcm:+" />
+        <framework src="com.google.firebase:firebase-messaging:9.0.0" />
```
macdonst: Should we make it 9+ to allow later versions of the lib?

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -58,8 +60,19 @@ public void setNotification(int notId, String message){
...
+
+        Bundle extras = new Bundle();
+
+        if (message.getNotification()!=null) {
+            extras.putString(TITLE,message.getNotification().getTitle());
+            extras.putString(MESSAGE,message.getNotification().getBody());
+        }
+        for (Map.Entry<String, String> entry : message.getData().entrySet()) {
+            extras.putString(entry.getKey(), entry.getValue());
+        }
```
macdonst: Hmmm....need to look into RemoteMessage some more but looks like a good idea to minimize changes for now.

src/android/com/adobe/phonegap/push/PushInstanceIDListenerService.java
```diff
@@ -5,23 +5,24 @@
...
-        SharedPreferences sharedPref = getApplicationContext().getSharedPreferences(COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
-        String senderID = sharedPref.getString(SENDER_ID, "");
-        if (!"".equals(senderID)) {
-            Intent intent = new Intent(this, RegistrationIntentService.class);
-                startService(intent);
-        }
+        // Get updated InstanceID token.
+        String refreshedToken = FirebaseInstanceId.getInstance().getToken();
+        Log.d(LOG_TAG, "Refreshed token: " + refreshedToken);
+        // TODO: Implement this method to send any registration to your app's servers.
```
macdonst: Here we will need to add:

```
 // save new token
SharedPreferences.Editor editor = sharedPreferences.edit();
editor.putString(REGISTRATION_ID, token);
editor.commit();
```

to save the new token or the PushPlugin.java code may try to load the old token. Also, we'll need to figure out a way to fire a `registration` event.

src/android/com/adobe/phonegap/push/PushPlugin.java
```diff
@@ -65,21 +65,7 @@ public void run() {
...
-                        }
-                        // new sender ID, re-register
-                        else if (!savedSenderID.equals(senderID)) {
-                            token = InstanceID.getInstance(getApplicationContext()).getToken(senderID, GCM);
-                        }
-                        // use the saved one
-                        else {
-                            token = sharedPref.getString(REGISTRATION_ID, "");
-                        }
+                        token = FirebaseInstanceId.getInstance().getToken();
```
macdonst: So if it never uses the SENDER_ID it must read it from the google_services.json file. We'll need to update the .init() JS code.

plugin.xml
matzew: @macdonst yes, the new google-services.json file contains that, and much more

src/android/com/adobe/phonegap/push/PushPlugin.java
matzew: @macdonst the "google-services.json" file needs to go into the native /app folder (root).

Not really sure how to get that mapping, from an app to the plugin

macdonst: @matzew yeah, we may need to do some "resource file" tricks or update cordova-android to be support this use case as I'm seeing it more and more.

secondsun: There are three options : 
 1) Apply the Project Number and Application IDs to firebase manually : https://gist.github.com/secondsun/dfd3a301a5cfbc34d74af18854276f1b
 2) Add to values.xml the following : gcm_defaultSenderId and google_app_id
 3) Include the plugin and the google-services.json file and let them do their magic.

matzew: 3) sounds more saine, staying w/ how it is supposed to work, no ?

secondsun: 3) is Google's "Golden Scenario"

jcesarmobile:
Rebased and removed some empty lines

src/android/com/adobe/phonegap/push/PushInstanceIDListenerService.java
jcesarmobile: I use `FirebaseInstanceId.getInstance().getToken()` to get the token, this should return the current token, so I don't think we have to store it on the SharedPreferences anymore

src/android/com/adobe/phonegap/push/PushPlugin.java
macdonst: Yeah, 3) would be better if we can deliver the file to the right location at compile time. If not we will have to go with 2) as plugin.xml can collect those values as variables and put them into values.xml.

macdonst:
Okay, merged into the v2.0.x branch. We can keep iterating on this. Please reference issue #929.

src/android/com/adobe/phonegap/push/PushPlugin.java
jcesarmobile: On my sample project I have this string on res/values/strings.xml (created by the plugin) `<string name="google_app_id">myAppId3333</string>` and I'm able to receive push notifications sent from FCM console, but I have nothing for gcm_defaultSenderId anywhere

secondsun: The senderID is part of the app id, so Firebase probably only loads that
value once.

Summers Pittman
>>Phone:404 941 4698
>>Java is my crack.

Issue #999, opened 2016-06-19
Is there any way to clear count badge in some Android devices?


oscarabilleira:
### Expected Behaviour
When push is recieved in some Android devices (Samsung, HTC...) is msgcount is included in playload a  count badge is added over the app icon, like in iois. Even if "clearBadge" option is included in the app (in android configuration line) when the app is opened the badge doesn't be cleared like in ios devices.
Is there any way to clear this badge when the user opens the app like in ios?

### Actual Behaviour
Badge is not cleared when the user opens the app.

### Reproduce Scenario (including but not limited to)
Send a push notification to android device (Samsung, Htc... ) that includes badge, 

#### Steps to Reproduce 
 Adding "msgcount:1" in play load. The badge apears over the icon but it is not cleared the opens the app.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Android 6 (but not only) in some devices.

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
Samsung, HTC...

#### Cordova CLI version and cordova platform version

    cordova --version                                6.1.0
    cordova platform version android                  5.1.1

#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   1.7.1

#### Sample Push Data Payload
Any playload that includes "msgcount":"1"

macdonst:
@oscarabilleira the `clearBadge` option is not implemented on Android but it is a good idea.

oscarabilleira:
Thanks for the answer...   I hope that in a near future this is included.

PR #1030, opened 2016-06-29
Implemented clearBadge for Android

Alex-Sessler:
## Description

I added the init parameter "clearBadge" which is treated the same as the others and stored to SharedPreferences. If it's set, the app sets the badge number to 0 on init. Also, when a notification is receviced, the badge number is immediately cleared in order to leave the config of the badge through notification parameter untouched. Now when a notification is recevied, the badge is cleared before the notification content is processed, therefore any badge numbers set in the notification will be respected. I tried to implement this as close to the iOS implementation as possible.

## Related Issue

https://github.com/phonegap/phonegap-plugin-push/issues/999

## Motivation and Context
Even though I was not setting the Badge Number, it was some times (on specific devices) appearing out of nowhere after init. I wanted to have a way to make sure that the Badge Number is not shown by accident.

## How Has This Been Tested?

I manually added a badge number through the "setApplicationBadgeNumber" function. Then I restarted the app causing the init function to trigger. The badge number was immediately cleared. I also set the BadgeNumber after init and sent a notification that had no badge number arguments. The badge number was also cleared immediately.

I only did manual tests on Android devices.

There shouldn't be any unexpected side effects on other code.
## Types of changes

## Checklist:


macdonst:
Thanks @Alex-Sessler

PR #1086, opened 2016-07-15
restart application after force close (#158 #333)

hanicker:
Restart of Android application if force closed (together with some istructions). Should solve #158 or #333.
Application started on background without blinking (at least with 2 tested devices running Android 5.0 and 6).


macdonst:
@hanicker thanks, I'm going to send a PR to add the necessary code to the `MainActivity` that cordova-android creates. I've been meaning to do that for awhile. Then when it is in I can merge your PR.

PR #1093, opened 2016-07-18
Added .framework files as custom frameworks

jcesarmobile:


fredgalvao:
Do you mind explaining what's the advantage of this change? I looked up the documentation on both, and I couldn't grasp the difference between <source-file> and <framework> other than the later being able to reference android frameworks without having to provide them (e.g. `<framework src="com.google.android.gms:play-services-gcm:+" />`)

jcesarmobile:
Basically I used `<source-file target-dir="FCM/Libraries" src="src/ios/Firebase/Analytics/FirebaseAnalytics.framework" framework="true" />` first and didn't work, then changed to `<framework src="src/ios/Firebase/Analytics/FirebaseAnalytics.framework" custom="true" />` and it worked.

This answer is just for iOS:
source-file should be used with libs (.a) and old custom static frameworks
framework should be used with system frameworks and custom dynamic frameworks.

macdonst:
I can't wait until CocoaPod support for plugins lands. The interns are working on it in SF.

PR #1103, opened 2016-07-20
Added hook and resource file to copy GoogleService-Info.plist

jcesarmobile:
Added a GoogleService-Info.plist resource file (empty) and a hook that copies the content of your GoogleService-Info.plist on the root of the project to the empty one on cordova prepare

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -118,6 +118,7 @@
         <framework src="src/ios/Firebase/Analytics/GoogleUtilities.framework" custom="true" />
         <framework src="src/ios/Firebase/Messaging/FirebaseMessaging.framework" custom="true" />
         <framework src="src/ios/Firebase/Messaging/GoogleIPhoneUtilities.framework" custom="true" />
+        <resource-file src="src/ios/Firebase/GoogleService-Info.plist" target="Resources/GoogleService-Info.plist" />
         <!--  End FCM Related Files -->
 	</platform>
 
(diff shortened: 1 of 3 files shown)


macdonst:
@jcesarmobile is it possible just to write the properties required by FCM in the apps plist? I try to avoid hooks whenever possible as they don't play well with cloud build services.

jcesarmobile:
The GoogleService-Info.plist file has to exist, if not the app will crash.
It has to contain values or it will crash too.
It has 17 keys-values, not sure if will crash if any of them is missing or just some of them.

Right now the resource-file element will add an empty GoogleService-Info.plist file, maybe we can use the config-file tag on the plugin.xml to fill the fields on GoogleService-Info.plist instead of doing it on the appName-info.plist, but I think 17 values are too much to ask an user to fill.

jcesarmobile:
I've just tested and app will only crash if GCM_SENDER_ID or GOOGLE_APP_ID are missing on the GoogleService-Info.plist, but the other values might be there for a reason, not sure how safe is not to add them.

macdonst:
@jcesarmobile so we might be able to pre-fill the 15 values that don't change and make the user provide the GCM_SENDER_ID & GOOGLE_APP_ID via a variable. Then we can write all 17 values uses the <config-file/> tag.

jcesarmobile:
No, most of them are custom values for your project. There are only 6 values that could be pre filled.

macdonst:
@jcesarmobile crap

jcesarmobile:
added a comment here https://github.com/cordova/cordova-discuss/issues/6

We need a resource-file tag on config.xml to copy this files

macdonst:
@jcesarmobile I agree 100%. We may have to take this upon ourselves.

PR #1169, opened 2016-08-19
Adding workshop tutorial link to the README

fredgalvao:
## Motivation and Context
Hoping someone will eventually read it, or the docs, :wink:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,4 +20,5 @@ This plugin offers support to receive and handle native push notifications with 
 
 # Do you like tutorial? You get tutorial!
 
- - [PhoneGap Day EU/US Push Workshop 2016](http://macdonst.github.io/push-workshop/)
+ - [PhoneGap Day US Push Workshop 2016 (using node-gcm)](http://macdonst.github.io/push-workshop/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,3 +16,8 @@ This plugin offers support to receive and handle native push notifications with
 - [Push notification payload details](docs/PAYLOAD.md)
 - [Contributing](.github/CONTRIBUTING.md)
 - [License (MIT)](MIT-LICENSE)
+
+
+# Do you like tutorial? You get tutorial!
+
+ - [PhoneGap Day EU/US Push Workshop 2016](http://macdonst.github.io/push-workshop/)
```
macdonst: There are actually two tutorials. The US one uses node-gcm/node-apns to send push data. The EU one uses the PhoneGap CLI http://macdonst.github.io/push-workshop-eu/, so I'd put in links to both.

fredgalvao: Interesting. I saw both links you posted, and I have them kept "at home" too, for a while now. But upon simple inspection, both use the same repo as a base, and thus I thought they were the same.

If they're both usefull each on it's own, I'll add both then with the info on where they differ.

macdonst: Yeah, they are mostly the same but the EU one is better for getting started as no need to setup certs/provs etc.

macdonst:
Thanks @fredgalvao

PR #1183, opened 2016-08-29
Update plugin to use GCM Cocoapods <framework> reference in plugin.xml

shazron:
## Description

This PR is to remove inclusion of GCM inside this plugin, and replace it Instead with a <framework> reference to the Cocoapod itself. A future version of the cordova-cli and the cordova-ios platform will support this new Cocoapod <framework> reference. 
## Motivation and Context

This change facilitates easy updates of GCM in this plugin.
## How Has This Been Tested?

Note that `pod setup` will clone a repo that is 475MB in size, so it might take a while.

**Test only with the cli nightlies starting with the nightly released on Monday Sept 5th 8PM PDT or install the `cordova-ios` platform from the git repo direct**

```
sudo gem install cocoapods
pod setup --verbose
npm install -g cordova@nightly
cordova create test test test
cd test
cordova platform add ios
cordova plugin add https://github.com/shazron/phonegap-plugin-push.git#gcm-cocoapod-support --variable SENDER_ID=xxxTestingxxx
cordova build
```

After testing, to revert to your original Cordova CLI or the latest:

```
npm install -g cordova@latest
OR
npm install -g cordova@VERSION
```

where VERSION is your specific version.
## Types of changes
## Checklist:

- [ ] I have updated the documentation accordingly ('with no dependency on any other plugins.' is not true anymore)
- [x] Add cordovaDependencies key in package.json http://cordova.apache.org/docs/en/latest/guide/hybrid/plugins/index.html#specifying-cordova-dependencies
- [x] Add engine element to plugin.xml http://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#engines-and-engine

This is a packaging change, no new tests need to be added.


jcesarmobile:
> A future version of the cordova-cli and the cordova-ios platform will support this new Cocoapod reference.

Shouldn't the plugin.xml have some property to only install in that future version?

shazron:
@jcesarmobile yes, see linked reference to the Roadmap issue https://github.com/phonegap/phonegap-roadmap/issues/131 where Steve has already commented on this. I didn't work on that yet since the implementation is incomplete

shazron:
Updated plugin.xml and package.json for dependencies. However, I have a problem when trying to install this plugin (this PR branch) into a project with cordova-cli 6.3.1:
```
Plugin doesn't support this project's cordova version. cordova: 6.3.1, failed version requirement: >=6.4.0
Skipping 'phonegap-plugin-push' for ios
```
It looks like it is honoring the `plugin.xml` `<engines>` tag, even though the `engines key` in `package.json` should be considered instead. Any insight here @stevengill ?

shazron:
Found the culprit. This [line](
https://github.com/apache/cordova-lib/blob/acb7b0a0e68499f589cf585830bfb6461d0e9f39/cordova-lib/src/cordova/plugin.js#L352).

id is `https://github.com/shazron/phonegap-plugin-push.git#gcm-cocoapod-support`
parsedSpec.version is `null`
cordova_util.isUrl(id) is `true`
cordova_util.isDirectory(maybeDir) is `false`.

Because second condition is true, that whole statement is true, so it returns the promise, and never gets to [getFetchVersion below](
https://github.com/apache/cordova-lib/blob/acb7b0a0e68499f589cf585830bfb6461d0e9f39/cordova-lib/src/cordova/plugin.js#L383).

macdonst:
@shazron is this PR good to be merged?

shazron:
Not yet, I'm waiting on @stevengill to comment on the above since we are not backwards compatible currently.

stevengill:
So check it, 

if you install the plugin via git or local path, `getFetchVersion` will not be called. It is only used when installing plugins via `npm`. 

So what this means is if you try to install a plugin that isn't supported by your platform/tools, you get a warning it isn't supported and it won't install. Users have to either use a older version of the plugin or update their projects to use the newer platforms/tools. This is expected behavior. If the user installs the plugin via npm, we will auto fetch the appropriate version based on their installed platforms/tools. 

I recommend not merging this change until cordova@6.4.0 and ios@4.3.0 are released. Right now, if people install the push plugin via git and you update master to this change, they will not be able to install and won't be able to update their existing projects since the new versions don't exist yet.

stevengill:
If you think most people are installing via npm, then go ahead and merge. Though this change isn't useable until ios@4.3.0 and cordova@6.4.0 are released.

shazron:
I'll defer to Simon on what he wants to do, to avoid any headaches.

Publish now:
- `npm` (cordova-cli < 6.1.0) users **will** have an upgrade path. 
- `npm` (cordova-cli >= 6.1.0) users **will** automatically downgrade to v1.8.1 
- `local` and `git url` users **will not** have an upgrade path (is this a significant no. of users?)
- we will have to document the second case, to install the plugin at version 1.8.1 explicitly

Publish once the respective cli and platforms are published:
- `npm` (cordova-cli < 6.1.0) users **will** have an upgrade path. 
- `npm` (cordova-cli >= 6.1.0) users **will** automatically downgrade to v1.8.1 
- `local` and `git url` users **will** have an upgrade path (upgrade platform, upgrade cli)
- no docs needed

macdonst:
@shazron why don't we just add information to the docs for local and git url users? It would go in:

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/INSTALLATION.md

shazron:
@macdonst sure, let's do that. So are you leaning to "publish now", then?

macdonst:
@shazron yeah, always be publishing. The earlier we get it out the quicker we get feedback and can improve on things.

shazron:
Ok, I'll update the docs for final review and will let all of you know when it is done.

shazron:
Ok docs updated for CocoaPods support. Have a look.

macdonst:
@shazron looks great. I'm going to merge on Monday as I don't want to do a release today. I'd prefer not to be fielding support questions all weekend.

wzoom:
Hi, is anything blocking the release still?

macdonst:
@wzoom I was traveling all last week so I did not get a chance to do a release of this functionality. It won't really be usable until cordova-ios 4.3.0 is release anyway so there is no big rush.

jcesarmobile:
You forgot to change the plugin version on plugin.xml

shazron:
@jcesarmobile thanks! updated.

shazron:
Rebased.

Issue #1194, opened 2016-09-05
Localization title and other constants on Android?


polyn0m:
My application uses 2 languages.

Is it possible to use the constants from strings.xml for notify title or summaryText? 

It is very expensive detect current language of user device on the server side.

macdonst:
@polyn0m currently this behaviour is not supported. It would be an interesting enhancement though.

Right now you best bet is to send the local of the user back to your server when you send it the registration ID. Then you can send separate push notifications out to your 2 languages.

polyn0m:
@macdonst tanks for quick reply.

Send locale to app server may cause errors, when user change language but not start app. It's rare case but possible. In my opinion store part of application on app server it is wrong.

I have start worked for this enhancement https://github.com/polyn0m/phonegap-plugin-push/commit/68c866363e8e0307e9334b43423c1644f380ba84 when finish it (hope 1-2 days), I'll make a pr.

PR #1196, opened 2016-09-06
Localization from resources

polyn0m:
## Description

This pull request added localization from resources strings.xml for title, message, summaryText.

## Related Issue

https://github.com/phonegap/phonegap-plugin-push/issues/1194

## Motivation and Context

This changes allows using localization from resources and not use app server for that.

## How Has This Been Tested?

This tested manually on existing application.

On screenshot used localized strings for title and summaryText.

loc_demo

## Types of changes

## Checklist:

Please review code for possible errors!

### DIFF
--- a/docs/PAYLOAD.md
+++ b/docs/PAYLOAD.md
@@ -101,8 +101,10 @@ var apiKey = "replace with API key";
 var deviceID = "my device id";
 var service = new gcm.Sender(apiKey);
 var message = new gcm.Message();
-message.addData('title', {"locKey": "push_app_title", "locData": ["custom data"]});
+message.addData('title', {"locKey": "push_app_title"});
 message.addData('message', 'Simple non-localizable text for message!');
+// Constant with formatted params
+// message.addData('message', {"locKey": "push_message_fox", "locData": ["fox", "dog"]});
 service.send(message, { registrationTokens: [ deviceID ] }, function (err, response) {
     if(err) console.error(err);
     else    console.log(response);
--- a/docs/PAYLOAD.md
+++ b/docs/PAYLOAD.md
@@ -3,6 +3,7 @@
    - [Background Events](#push-message-arrives-with-app-in-background)
    - [Tap Events](#user-clicks-on-notification-in-notification-center)
 - [Android Behaviour](#android-behaviour)
+  - [Localization](#localization)
   - [Images](#images)
   - [Sound](#sound)
   - [Stacking](#stacking)
@@ -62,6 +63,59 @@ Some ways to handle this *double* event are:
 - include a unique ID in your push so you can check to see if you've already processed this event.
 
 # Android Behaviour
+
+## Localization
+
+Plugin supported localization from resources for: title, message and summaryText.
+
+You may use simple link to locale constant.
+
+```javascript
+{
+    "registration_ids": ["my device id"],
+    "data": {
+        "title": {"locKey": "push_app_title"},
+        "message": "Simple non-localizable text for message!"
+    }
+}
+```
+
+Or use localization with formatted constants.
+
+```javascript
+{
+    "registration_ids": ["my device id"],
+    "data": {
+        "title": {"locKey": "push_app_title"},
+        "message": {"locKey": "push_message_fox", "locData": ["fox", "dog"]}
+    }
+}
+```
+
+Here is an example using node-gcm that sends the above JSON:
+
+```javascript
+var gcm = require('node-gcm');
+// Replace these with your own values.
+var apiKey = "replace with API key";
+var deviceID = "my device id";
+var service = new gcm.Sender(apiKey);
+var message = new gcm.Message();
+message.addData('title', {"locKey": "push_app_title", "locData": ["custom data"]});
+message.addData('message', 'Simple non-localizable text for message!');
+service.send(message, { registrationTokens: [ deviceID ] }, function (err, response) {
+    if(err) console.error(err);
+    else    console.log(response);
+});
+```
+
+Localization must store in strings.xml
+
+```xml
+<string name="push_app_title">@string/app_name</string>
+<string name="push_message_fox">The quick brown %1$s jumps over the lazy %2$s</string>
+<string name="push_summary_text">%%n%% new message(s)</string>
+```
 
 ## Images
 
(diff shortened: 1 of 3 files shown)


docs/PAYLOAD.md
```diff
@@ -63,6 +64,59 @@ Some ways to handle this *double* event are:
...
+Here is an example using node-gcm that sends the above JSON:
+
+```javascript
+var gcm = require('node-gcm');
+// Replace these with your own values.
+var apiKey = "replace with API key";
+var deviceID = "my device id";
+var service = new gcm.Sender(apiKey);
+var message = new gcm.Message();
+message.addData('title', {"locKey": "push_app_title", "locData": ["custom data"]});
```
macdonst: I'd remove the `locData` from this example as `push_app_title` does not support variables.

docs/PAYLOAD.md
```diff
@@ -63,6 +64,59 @@ Some ways to handle this *double* event are:
...
+message.addData('message', 'Simple non-localizable text for message!');
+service.send(message, { registrationTokens: [ deviceID ] }, function (err, response) {
```
macdonst: Maybe add line showing how to send a `message` with variables. It can even be commented out.

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -119,6 +122,49 @@ private void replaceKey(String oldKey, String newKey, Bundle extras, Bundle newE
     }
 
     /*
+     * Normalize localization for key
+     */
+    private String localizeKey(Context context, String key, String value) {
+        if (key.equals(TITLE) || key.equals(MESSAGE) || key.equals(SUMMARY_TEXT)) {
+            try {
+                JSONObject localeObject = new JSONObject(value);
```
macdonst: Instead of just trying to create a JSONObject from the string passed in can we do a check to see if it needs to be localized? That way for apps not localizing their push messages they won't throw 2 JSONExceptions for every push message received. Something like:

```
if (value.toLowerCase().contains("locKey".toLowerCase()) {
  // do stuff
}
```

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -119,6 +122,49 @@ private void replaceKey(String oldKey, String newKey, Bundle extras, Bundle newE
...
+
+                String localeKey = localeObject.getString("locKey");
```
macdonst: Make `locKey` a constant.

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -119,6 +122,49 @@ private void replaceKey(String oldKey, String newKey, Bundle extras, Bundle newE
...
+                
+                ArrayList<String> localeFormatData = new ArrayList<String>();
+                if (!localeObject.isNull("locData")) {
```
macdonst: Make `locData` a constant.

docs/PAYLOAD.md
```diff
@@ -63,6 +64,59 @@ Some ways to handle this *double* event are:
...
+    if(err) console.error(err);
+    else    console.log(response);
+});
+```
+
+Localization must store in strings.xml
```
macdonst: Can we explain to users how to add strings to strings.xml?

macdonst:
@polyn0m looks pretty good, I added some code comments for you to respond to.

src/android/com/adobe/phonegap/push/GCMIntentService.java
fredgalvao: I disagree. It's a protective approach to not try to do string search out of context, as a full payload could have the string "lockey" somewhere without being necessarily a json key.

docs/PAYLOAD.md
fredgalvao: Let's try to use config.xml (edit-config?) as much as possible so we don't "break" the cli-centric workflow for cordova users by making them modify output files (`/platforms/android/`)

src/android/com/adobe/phonegap/push/GCMIntentService.java
macdonst: I agree with my earlier comment and @fredgalvao. I said `something like` because I'd like to get rid of the JSONExceptions.

polyn0m: Maybe use this (similar method [.mayBeJSON(String str)](http://json-lib.sourceforge.net/apidocs/jdk15/net/sf/json/util/JSONUtils.html#mayBeJSON%28java.lang.String%29)  available in the JSONUtils library):

```java
if (value.startWith("{") && value.endWith("}")
    && value.toLowerCase().contains("locKey".toLowerCase()) {
  // do stuff
}
```

I doubt that users show the string containing this, but may be exceptions.

docs/PAYLOAD.md
polyn0m: @fredgalvao Cordova config.xml? How it may use for edit strings.xml? I know this possible in plugin.xml, but this method not allow use different constant names.

polyn0m: @macdonst may be like this:

You may added constants to `platforms/android/res/values/strings[-LANGUAGE].xml` manually ([Android strings](https://developer.android.com/guide/topics/resources/string-resource.html)), but if you remove or re-add platform it may drop all you changes. Or you can use hooks for add constants automatically. 

IMPORTANT: Plugin supports only string values.

fredgalvao: @polyn0m Check [this release post](https://cordova.apache.org/news/2016/07/13/tools-release.html) for details on the `<edit-config>` tag.

Also, kinda deviated from the topic, but please @macdonst clarify something for me: will the `<edit-config>` tag be available for the user on the project's `config.xml`, or is it a plugin-only feature? If it's the later, I'll be surprisingly disappointed this wasn't made available to users.

src/android/com/adobe/phonegap/push/GCMIntentService.java
fredgalvao: That minimzes the chances of a miss-catch, but it still exists. If we don't want to deal with the exceptions and can't use
```xml
<groovy-json-on-android-dream> a diff json lib </groovy-json-on-android-dream>
```
then this would be the the closest to safety, I guess.

docs/PAYLOAD.md
macdonst: @fredgalvao `edit-config` will be available to config.xml.

polyn0m:
@macdonst some small fixes, but fixes about exceptions and docs for strings.xml not added. Please comment need these fixes or not.

macdonst:
@polyn0m thanks, merged and 1.8.2 being released shortly.

@fredgalvao we can iterate on the docs.

PR #1225, opened 2016-09-18
Use unique pending intent request code to enable multiple...

mikepsinn:
…simultaneous notifications with action buttons

## Description
Use a unique pending intent id by appending i to the notId for each action button for each Android notification (instead of notId + 0, 1, 2).

## Related Issue
https://github.com/phonegap/phonegap-plugin-push/issues/1215

## Motivation and Context
This allows for multiple simultaneous Android notifications with action buttons. Previously, all buttons were given pending intend request code notId + i. However, this doesn't work if notId's are sequential. 

## How Has This Been Tested?
Manually and with npm test.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] I have added tests to cover my changes. (I don't really see any Android specific tests)

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -430,7 +430,8 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
                 JSONArray actionsArray = new JSONArray(actions);
                 ArrayList<NotificationCompat.Action> wActions = new ArrayList<NotificationCompat.Action>();
                 for (int i=0; i < actionsArray.length(); i++) {
-                    int uniquePendingIntentRequestCode = i + notId;
+                    // This effectively appends i's value to the notId to ensure uniqueness when multiple notId's are sequential
+                    int uniquePendingIntentRequestCode = notId * 10 + i;
                     Log.d(LOG_TAG, "adding action");
                     JSONObject action = actionsArray.getJSONObject(i);
                     Log.d(LOG_TAG, "adding callback = " + action.getString(CALLBACK));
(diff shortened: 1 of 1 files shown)


src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -430,7 +430,8 @@ private void createActions(Bundle extras, NotificationCompat.Builder mBuilder, R
                 JSONArray actionsArray = new JSONArray(actions);
                 ArrayList<NotificationCompat.Action> wActions = new ArrayList<NotificationCompat.Action>();
                 for (int i=0; i < actionsArray.length(); i++) {
-                    int uniquePendingIntentRequestCode = i + notId;
+                    // This effectively appends i's value to the notId to ensure uniqueness when multiple notId's are sequential
+                    int uniquePendingIntentRequestCode = notId * 10 + i;
```
macdonst: @mikepsinn could we get an integer overflow problem here? Why the need to multiply by 10?

mikepsinn: @macdonst Yeah, I guess you could if your notId is greater than 214,748,364.  

I had to do it because I had a two notId's that were 100 and 101 so the codes were 100, 101, 102 and 101, 102, 103, respectively.  If I multiply by 10 with my change, they'll be 1000, 1001, 1002 and 1010, 1011, and 1012.   Note those in the second group are all unique. 

Should I just generate random integers like you originally suggested?

mikepsinn:
@macdonst I just added another commit that just assigns random numbers instead. It seems to work fine as long as I don't try it two billion times.

macdonst:
Yeah, I think that is the best way to go.

PR #1245, opened 2016-09-27
Add APNS Environment information (Compiles under XCode 8)

jimduchek:
## Description
This is identical to jcaron23's push request, fixed for XCode 8

## Motivation and Context
Allows apps to support both production and sandbox environments.

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [x ] My code follows the code style of this project.
- [x ] I have updated the documentation accordingly.
- [ x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -157,6 +157,9 @@ The event `registration` will be triggered on each successful registration with 
 Parameter | Type | Description
 --------- | ---- | -----------
 `data.registrationId` | `string` | The registration ID provided by the 3rd party remote push service.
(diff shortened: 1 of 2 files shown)


macdonst:
@jimduchek so why did you close this PR? Does it not work correctly?

jimduchek:
It does not, but it's possibly minor:  The aps-environment string is indeed 'production' in an app installed with an ad-hoc production cert from itunes, but it is... something other than 'production' in an app that comes through the app store.  I do not know what it is, as I can't seem to get at the console in a proper app-store installed app -- it may be as simple as being ' production' or the like.  I changed my code for my app to instead query against the string 'development' and it works for now, but I pulled the PR until I could verify.

macdonst:
@jimduchek okay, when you get a chance to verify please re-send the PR.

Issue #1265, opened 2016-10-05
[Fixed] iOS 10.0.1 - Plugin 1.8.2 - OnNotification event no more fired when app is in background


thsantac:
### Expected Behaviour

When the app is in background and a notification is received, app should wake up and do some actions based on the onNotification event.
### Actual Behaviour

App wakes up but doesn't do expected actions. If app is killed before receiving notification, expected actions are done.
### Reproduce Scenario (including but not limited to)

Put app in background and send a notification.
#### Steps to Reproduce

Put app in background and send a notification.
#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)

iOS 10.0.1
#### Cordova CLI version and cordova platform version

```
cordova --version                                    6.2.0
cordova platform version android           android 5.1.1 - ios 4.1.1
```
#### Plugin version

```
cordova plugin version | grep phonegap-plugin-push   # e.g. 1.8.2
```
#### Sample Push Data Payload

Not available (WordPress notification plug-in)

macdonst:
@thsantac Sorry, I need more information in order to debug this. When a notification is received on iOS the `on('notification')` event handler will not be called unless you are sending the property in your push payload to indicate it is a background message.

So, I'm a little confused as to what you are doing. Is it a background message, are you tapping the notification, etc? More information will help and the push payload is the key piece.

thsantac:
All I have is the WordPress plug-in log:

INFO: Message : Array ( [MESSAGE] => ApnsPHP_Message Object ( [_bAutoAdjustLongPayload:protected] => 1 [_aDeviceTokens:protected] => Array ( [0] => 06b7c606722c76693c31b70e28feb406a152baa836135b1f3663df4b10d19ee3 ) [_sText:protected] => Postes ouverts à la CAP de décembre 2016 [_nBadge:protected] => 2 [_sSound:protected] => default [_sCategory:protected] => [_bContentAvailable:protected] => 1 [_aCustomProperties:protected] => Array ( [id] => 11833 ) [_nExpiryValue:protected] => 604800 [_mCustomIdentifier:protected] => ) [BINARY_NOTIFICATION] => 	W?5 ??r,vi<1?(???R??6[6c?K???{**"aps":{"alert":"Postes ouverts à la CAP de décembre 2016","badge":2,"sound":"default","content-available":1,"priority":"high"},"id":"11833"**} [ERRORS] => Array ( ) )

Is there a missing property I should add?

[...]

PR #1279, opened 2016-10-12
1. Using UNUserNotificationCenterDelegate (required to know the app s…

Lucklyric:
Using UNUserNotificationCenter to handle push notification in iOS10
## Description

application:didReceiveRemoteNotification does not properly detect the app's state is iOS10
and not fired after app wakes up from background by tapping a notification.
## Related Issue
#1265
## How Has This Been Tested?

I tested on my own iOS 10 phone and iOS 9 iPad.
## Types of changes
- [ X] Bug fix (non-breaking change which fixes an issue)
## Checklist:
- [X ] My code follows the code style of this project.
- [ X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -1,16 +1,16 @@
 /*
  Copyright 2009-2011 Urban Airship Inc. All rights reserved.
- 
+
  Redistribution and use in source and binary forms, with or without
  modification, are permitted provided that the following conditions are met:
- 
+
  1. Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.
- 
+
  2. Redistributions in binaryform must reproduce the above copyright notice,
  this list of conditions and the following disclaimer in the documentation
  and/or other materials provided withthe distribution.
- 
+
  THIS SOFTWARE IS PROVIDED BY THE URBAN AIRSHIP INC``AS IS'' AND ANY EXPRESS OR
  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
@@ -181,7 +181,7 @@
             UserNotificationTypes |= UIUserNotificationTypeBadge;
 #endif
         }
-        
+
         if (([soundArg isKindOfClass:[NSString class]] && [soundArg isEqualToString:@"true"]) || [soundArg boolValue])
         {
             notificationTypes |= UIRemoteNotificationTypeSound;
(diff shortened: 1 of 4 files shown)


src/ios/PushPlugin.m
```diff
@@ -619,4 +643,20 @@ -(void)stopBackgroundTask:(NSTimer*)timer
     }
 }
 
+//- (void)userNotificationCenter:(UNUserNotificationCenter *)center
```
macdonst: Can you remove all this commented out code?

src/ios/PushPlugin.m
```diff
@@ -578,25 +600,27 @@ -(void)failWithMessage:(NSString *)message withError:(NSError *)error
...
 -(void) finish:(CDVInvokedUrlCommand*)command
 {
     NSLog(@"Push Plugin finish called");
-
+    
     [self.commandDelegate runInBackground:^ {
         NSString* notId = [command.arguments objectAtIndex:0];
-
+        
+        UIApplication *app = [UIApplication sharedApplication];
```
macdonst: This seems to be an unused variable. Were you intending to do something with UIApplication or can it be removed?

src/ios/PushPlugin.m
```diff
@@ -1,16 +1,16 @@
 /*
  Copyright 2009-2011 Urban Airship Inc. All rights reserved.
-
+
```
macdonst: Can you fix all of the formatting problems with the file? It will make the review a lot easier.

src/ios/PushPlugin.m
Lucklyric: Yes it is, I have removed it.

src/ios/PushPlugin.m
Lucklyric: Removed it.

Lucklyric:
It seems not fix @thsantac 's problem. Alought it works for me, I will still keep him in touch to make sure these changes is working.

macdonst:
@Lucklyric awesome, thanks for making those changes. I'm juggling some stuff but hope to be able to test/merge sometime tomorrow.

macdonst:
@Lucklyric here are the tests cases I run through:

1) ✅ When I send a message with the app in the foreground I get the event.
2) ✅ When the app is in the background I send a message and when I click on it in the notification shade the event is fired.
3) ❌ When the app is in the background I send a message with content-available: 1 and when I open the app the notification event has been fired.
4) ✅ When I kill the app, I send a message and when I click on it in the notification shade the event is fired.
5) ❌ When I kill the app, I send a message with content-available: 1 and when I open the app the notification event has not been fired. This is kinda expected behaviour considering there is not web view to run the JS.

The failure of 5 is to be expected but 3 is a regression. When I send a background push I see the following log in XCode even though I'm calling `push.finish()`.

```
Warning: Application delegate received call to -application:didReceiveRemoteNotification:fetchCompletionHandler: but the completion handler was never called.
```

Without fixing that we can have real problems with iOS apps.

appzer:
Hi,

i tested this pr on my iphone6 with iOS 10.1.1

The action Buttons are registered correctly but never called after pressing one.

Kevin

gustavokloh:
How is this pull request going?

macdonst:
Closing in favour of #2363

PR #1292, opened 2016-10-14
Remove warnings about unused variables and empty lines

dannywillems:


jcesarmobile:
Thanks for the PR, but instead of removing sendDataMessageFailure and sendDataMessageSuccess lines, I think we should pass that information to the javascript side so the developer can use it.

PR #1396, opened 2016-11-22
fix: avoid double [FIRApp configure] (crash)

pwbs:
## Description

I've quite naively applied what I found in http://stackoverflow.com/questions/38536257/ios-extension-fatal-exception-com-firebase-core-default-app-has-already-been

**My apologies for not having time to read your guidelines.**

## Motivation and Context

This patch is very trivial. 

## How Has This Been Tested?

It just stopped crashing (SIGABRT)...

## Types of changes

## Checklist:

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -291,7 +291,8 @@
             NSLog(@"Using FCM Notification");
             [self setUsesFCM: YES];
             dispatch_async(dispatch_get_main_queue(), ^{
-                [FIRApp configure];
+                if([FIRApp defaultApp] == nil)
+                    [FIRApp configure];
                 [self initRegistration];
             });
         } else {


jcesarmobile:
Looks good, but how are you having this issue? do you call init multiple times?

pwbs:
> Looks good, but how are you having this issue? do you call init multiple times?

When someone authenticates, or signs out, we restart the app. Each time the app is restarted, we call the initialisation function. It seems that we're doing it wrong...
**So, I don't know if this PR is relevant.** I'm using the plugin this way for now, I think we should review our initialisation process. 

I'm very new to this plugin, I don't know much about how it works yet, I just thought that perhaps it could be helpful to other people.

fredgalvao:
A raw reload of the html files on cordova, although rare and not advised in SPA, is a case and we should handle it, so this seems like a valid change for me.

macdonst:
LGTM too but @jcesarmobile and I are in the process of rebasing v2.0.x to be applied on top of master so we get everything that 1.9.0 has then we should merge this.

PR #1416, opened 2016-11-28
Update Plugin.xml to reference latest firebase-messaging framework

kelvinhokk:
Update framework for com.google.firebase to use + in gradle config.

## Description

Plugin.xml framework now references com.google.firebase:firebase-messaging:+ instead of com.google.firebase:firebase-messaging:9+

## Related Issue

To reproduce existing issue:
- start a new cordova project
- install version 1.7.2 of cordova-plugin-google-analytics
- install latest 2.0.x branch of phonegap-plugin-push commit 05b8f78
- run cordova run android.

## Motivation and Context

The plugin.xml framework src now references com.google.firebase:firebase-messaging:9+.  This conflicts with other plugins like cordova-plugin-push-analytics which uses com.google.android.gms:play-services-analytics:+

In addition, the documentation for [firebase]( https://firebase.google.com/docs/android/setup) recommends to use Google Play Services 10.0.0 and up.

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ X] My code follows the code style of this project.
- [ X] I have read the **CONTRIBUTING** document.
- [X ] All new and existing tests passed.


macdonst:
@kelvinhokk please rebase your branch so we don't end up pulling in 15 commits.

plugin.xml
```diff
@@ -1,123 +1,122 @@
...
+                <intent-filter>
+                    <action android:name="com.google.firebase.INSTANCE_ID_EVENT" />
+                </intent-filter>
+            </service>
+		</config-file>
+
+        <framework src="push.gradle" custom="true" type="gradleReference" />
+        <framework src="com.android.support:support-v13:23+" />
+        <framework src="me.leolin:ShortcutBadger:1.1.11@aar"/>
+        <framework src="com.google.firebase:firebase-messaging:+" />
```
macdonst: What is the rational behind switching this from `9+` to just `+`?

macdonst: Yeah, I don't really like this as + means any version of the FCM libs so we could get some weird behaviour.

fredgalvao: Is this an argument against the `loose` aspect of `:+`, or against version 10 itself? If it's the former, we can just switch from 9+ to 10+. If the later, is there any known misbehaviour coming from v10 so far?

kelvinhokk: i've done a quick scan on the plugins on ionic native and other plugins that references the google play services and this is what I found:

cordova-plugin-googleplus - ionic native
```
<framework src="com.google.android.gms:play-services-auth:+" />
<framework src="com.google.android.gms:play-services-identity:+" />      
```
cordova-plugin-googlemaps - ionic native
```
<framework src="com.google.android.gms:play-services-maps:9.8.0" />
<framework src="com.google.android.gms:play-services-location:9.8.0" />
```
cordova-plugin-google-analytics - ionic native
```
<framework src="com.google.android.gms:play-services-analytics:+" />
```

cordova-admob-pro - ionic native
```
<framework src="com.google.android.gms:play-services-ads:+" />
```

cordova-plugin-analytics
```
<framework src="com.google.android.gms:play-services-analytics:+" />
```

cordova-admob
```
<framework src="com.google.android.gms:play-services-ads:+" />
```

cordova-plugin-firebase
```<framework src="com.google.firebase:firebase-core:+" />
<framework src="com.google.firebase:firebase-messaging:+" />
<framework src="com.google.firebase:firebase-crash:+" />
<framework src="com.google.firebase:firebase-config:+" />
```

I understand even Android Studio advises again using the "+" in version numbers as it may lead to unpredicatable builds,  but then using "+" for now seems to be the only way to maintain compatibility with the other plugins out there.

jcesarmobile: Maybe we should open an issue on issues.cordova.io to make it use the more restrictive in case there are two plugins with the same dependency instead of conflicting. @macdonst WDYT?

@fredgalvao there is no 10 yet, but we can't assume the plugin will work when it's released, as it's a major version bump, it will probably break things

macdonst:
Not going to merge this in. Having some insane problems with Gradle and FCM. Applying the `com.google.gms.google-services` plugin only works if you specify an exact version of FCM.

Issue #1446, opened 2016-12-08
Unable to send inline reply from notification action button


bmwertman:
### Expected Behaviour
Select the `REPLY` button and have the inline reply field displayed in the notification interface.

### Actual Behaviour
Upon selecting the `REPLY` action button application is brought from the background to the foreground.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
     Android 5.0.2

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
    Oneplus One

#### Cordova CLI version and cordova platform version
    cordova --version                          6.4
    cordova platform version android           6.1
    cordova platform version ios           4.3

#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   1.9.1

#### Sample Push Data Payload
```
{
    priority: "high",
    data: {
            "content-available": "1",
            title: req.body.sender_name,
            body: req.body.message,
            state: req.body.state,
            roomId: req.body.roomId,
            ledColor: [17, 2, 255, 1],
            soundname: "default",
            image: req.body.sender_imgURL,
            sender_imgUrl: req.body.sender_imgURL,
            actions: [
              {
                "icon": "chat",
                "title": "Chat",
                "foreground": true,
                "callback":"showChatter"
              },
              {
                "icon": "reply",
                "title": "Reply",
                "inline": true,
                "foreground": false,
                "callback": "inlineReply"
              }
            ]
          }
  }
```
#### Sample Code that illustrates the problem
app.js inside `$ionicPlatform.ready`
```
  var launchChat = true;
    window.inlineReply = function(){
      launchChat = false;
      console.log("User replied inline");
    }

    window.showChatter = function(){
      console.log("Taking you to the chat");
    }

    // triggered every time notification received
    $rootScope.$on('$cordovaPushV5:notificationReceived',function(event, notification){
      var data = notification.additionalData;
      if(launchChat){
        if(data.sender_imgUrl !== userService.getUser().pic && data.foreground){// Make sure the user isn't getting notified about their own message
          //Notification was received in foreground. Check if the user is already in the room
          if(!$state.is(data.state, { roomId: data.roomId })){
            var queue = document.getElementById('queue');
            if(!queue){ // This is the first queued chatter
              var body = angular.element(document.getElementsByTagName('body')[0]),
                  messageCount = angular.element(document.createElement('span')),
                  imgTag = angular.element(document.createElement('img')),
                  remove = angular.element(document.createElement('div')),
[...]

PR #1459, opened 2016-12-13
Ensures foreground is true when inline is set to true and Android version is earlier than N  

bmwertman:
## Description
In `PushHandlerActivity.Java` I added a condition if `inline` is set to `true` but the client Android version was below N then the `foreground` should always be set to `true`. 

## Related Issue
#1446

## Motivation and Context
When trying to add inline reply to support Android N and later, clients running earlier versions of Android would not have their inline action callback ran. 

This left the notification's action button appearing as it did nothing.

## How Has This Been Tested?
Checked to ensure app handled notifications from closed, background and foreground as expected. App was launched to foreground when in the closed or background state. 
Tested on the following device & environment;
#### Platform and Version
     Android 5.0.2

#### What device vendor 
    Oneplus One

#### Cordova CLI version and cordova platform version
    cordova --version                          6.4
    cordova platform version android           6.1
    cordova platform version ios           4.3

#### Plugin version
    cordova plugin version | grep phonegap-plugin-push   1.9.1

I don't have a Android device running Android N to test on. So no tests have been ran for that version.

## Types of changes

## Checklist:

### DIFF
--- a/docs/PAYLOAD.md
+++ b/docs/PAYLOAD.md
@@ -606,7 +606,7 @@ This will produce the following notification in your tray:
 
 ![action_combo](https://cloud.githubusercontent.com/assets/353180/9313435/02554d2a-44f1-11e5-8cd9-0aadd1e02b18.png)
 
-If your users clicks on the main body of the notification your app will be opened. However if they click on either of the action buttons the app will open (or start) and the specified JavaScript callback will be executed. In this case it is `app.emailGuests` and `app.snooze` respectively. If you set the `foreground` property to `true` the app will be brought to the front, if `foreground` is `false` then the callback is run without the app being brought to the foreground.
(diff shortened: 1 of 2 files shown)


fredgalvao:
Thanks for the PR!

> I wasn't sure if this needed a documentation update. Let me know if there is anything else I can do to make this a better PR. 

We might as well update/improve [this section of the docs](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/PAYLOAD.md#in-line-replies), to make it clear that these two sections merge on the conditionals:

> On Android M and earlier the action buttons will work exactly the same as before but on Android N and greater when the user clicks on the Email Guests button you will see the following:

> Then your app's on('notification') event handler will be called without the app being brought to the foreground and the event data would be:

macdonst:
@bmwertman thanks for the bug fix. I've tested and checked it into master. I'll release a 1.9.2 soon. Also, next time I'm in NYC I owe you a coffee.

bmwertman:
Not a problem. You've helped me on more than a couple issues. I'll shoot you an email with my phone. Definitely give me a ring if you're in town.

Issue #1464, opened 2016-12-14
Create round bitmap icon for large icon image passed in from local resource or url


bmwertman:
I wanted the large icon in the push notification on Android to match the default Android style of being round instead of square.

I plugged the `getCircleBitmap` method shown in [this blog](http://curious-blog.blogspot.de/2014/05/create-circle-bitmap-in-android.html) into the GCMIntentService.java file. I've tested it on my Android L device and it converts the square icon to round with no issues.

I would like to create a pull request for this but because it is really a feature the PR prompts suggested creating an issue first.

fredgalvao:
I think that's great, and maybe it can live behind a setting specific for android, passed on `init`. Also, it's important to double check if that compiles with lower android target SDKs.

bmwertman:
I searched around to see when the large icon for push notifications became a standard feature on Android but as far as I could tell it has always been that way, at least back to Android 4. @fredgalvao  what is the best way to be certain that this will compile down to lower target SDK's?

As you will see in my commits to the PR I also initially thought that it should be a configurable setting. Although I tried to set it up as a property passed on `data` in the payload instead of on `init`. Something like `roundIcon: true`. 

I was having trouble getting a conditional setup properly to create the desired effect though. And the more I thought about it I couldn't think of a case where you wouldn't want the icon to match the others in the notification tray. So I'm a little more for just making it the default style.

Should I go ahead and make the PR for this and link back to this issue?

fredgalvao:
Well, if it has been the standard since `sdk=14`, then I'm all for making it the default behaviour. I still am not sure if we should make it the **only** option.

As for the **compile** thing, it's just a good habit to try and compile it with a lower target sdk (like 14, if we want to make the previous point literal). I have seen situations where a new API was used on a project/plugin, but it wouldn't compile on lower versions, and the developer forgot to use the compat library for that, thus breaking the project. Basically, that means assuring that all classes/objects used on that new code of yours are present and available uniformly throughout all supported versions of the android sdk.

Considering both points above, I think we're/we'll be ready for the PR.

I'm now kinda surprised me never seeing my app icon as round as strange...

bmwertman:
Sounds reasonable. For reference here is the method added in `GCMIntentService.java`. It only uses a few additional Classes from `android.graphics`. The [developer reference](https://developer.android.com/reference/android/graphics/package-summary.html) says they were  introduced in the level 1 API. Here is the method for reference;

```
[...]

PR #1468, opened 2016-12-15
Adds getCircleBitmap method in GCMIntentService.java to be used in setNotificationLargeIcon

bmwertman:
## Description
Adds method `getCircleBitmap` and calls it within `setNotificationLargeIcon` passing in either the bitmap created from image URL or the image retrieved from local assets. Returns a round bitmap.

## Related Issue
#1464 

## Motivation and Context
Large icons in Android push notifications are by default round. The plugin prior to introducing the `getCircleBitmap` method created a square icon. `getCircleBitmap` converts all large icon images to a round bitmap prior to inserting them keeping with the default large icon style. 

## How Has This Been Tested?
Checked to ensure app handled notifications from closed, background and foreground as expected.
Tested on the following device & environment;
#### Platform and Version
     Android 5.0.2

#### What device vendor 
    Oneplus One

#### Cordova CLI version and cordova platform version
    cordova --version                          6.4
    cordova platform version android           6.1
    cordova platform version ios           4.3

#### Plugin version
    cordova plugin version | grep phonegap-plugin-push   1.9.1 with pulled changes from 1.9.2

## Types of changes

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -13,6 +13,7 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
+import android.graphics.PorterDuff;
 import android.graphics.PorterDuffXfermode;
 import android.graphics.Rect;
 import android.graphics.RectF;
@@ -717,8 +718,8 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
     }
 
     private void setNotificationLargeIcon(Bundle extras, String packageName, Resources resources, NotificationCompat.Builder mBuilder) {
-        String gcmLargeIcon = extras.getString(IMG_SHAPE); // from gcm
-        String gcmLargeIconShape = extras.getString(SHAPE); // from gcm
+        String gcmLargeIcon = extras.getString(IMAGE); // from gcm
+        String gcmLargeIconShape = extras.getString(IMG_SHAPE); // from gcm
         if (gcmLargeIcon != null && !"".equals(gcmLargeIcon)) {
             if (gcmLargeIcon.startsWith("http://") || gcmLargeIcon.startsWith("https://")) {
                 Bitmap bitmap = getBitmapFromURL(gcmLargeIcon);
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -719,7 +719,7 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
 
     private void setNotificationLargeIcon(Bundle extras, String packageName, Resources resources, NotificationCompat.Builder mBuilder) {
         String gcmLargeIcon = extras.getString(IMAGE); // from gcm
-        boolean gcmLargeIconRound = extras.getBoolean(ROUND_IMG, false); // from gcm
+        boolean gcmLargeIconRound = extras.getBoolean(ROUND_IMAGE, false); // from gcm
         if (gcmLargeIcon != null && !"".equals(gcmLargeIcon)) {
             if (gcmLargeIcon.startsWith("http://") || gcmLargeIcon.startsWith("https://")) {
                 Bitmap bitmap = getBitmapFromURL(gcmLargeIcon);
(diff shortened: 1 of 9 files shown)


macdonst:
@bmwertman no, circular icons are not the default. By convention if you are showing a picture of a person google tends to use a circle but for other images it recommends rectangles. You can see that in their material design guidelines:

https://material.io/guidelines/patterns/notifications.html#

Not to say this isn't a good bonus feature for the plugin.

However, you mention you copied the code from a blog post. Did you get permission from the blog post owner to use that code? If the code does not have a license associated with it then it is assumed to be copywritten by the author of the post. I don't want to run into any legal issues.

Finally, I don't like the fact that it is always a circular icon. That should be an option.

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -690,19 +696,43 @@ private void setNotificationPriority(Bundle extras, NotificationCompat.Builder m
...
+
     private void setNotificationLargeIcon(Bundle extras, String packageName, Resources resources, NotificationCompat.Builder mBuilder) {
         String gcmLargeIcon = extras.getString(IMAGE); // from gcm
         if (gcmLargeIcon != null && !"".equals(gcmLargeIcon)) {
             if (gcmLargeIcon.startsWith("http://") || gcmLargeIcon.startsWith("https://")) {
-                mBuilder.setLargeIcon(getBitmapFromURL(gcmLargeIcon));
+                Bitmap bitmap = getBitmapFromURL(gcmLargeIcon);
+                Bitmap bm = getCircleBitmap(bitmap);
+                mBuilder.setLargeIcon(bm);
                 Log.d(LOG_TAG, "using remote large-icon from gcm");
```
macdonst: I'd like to see the above code be something like this, right above the if statement add:

```
boolean roundIcon = extras.getBoolean(ROUND_ICON):
```

of course `ROUND_ICON` will need to be added to the constants file.

then inside the if/else you could do:

```
if (roundIcon) {
  mBuilder.setLargeIcon(getCircleBitmap(getBitmapFromURL(gcmLargeIcon)));
} else {
  mBuilder.setLargeIcon(getBitmapFromURL(gcmLargeIcon)); 
}
```

src/android/com/adobe/phonegap/push/GCMIntentService.java
```diff
@@ -690,19 +696,43 @@ private void setNotificationPriority(Bundle extras, NotificationCompat.Builder m
...
             } else {
                 AssetManager assetManager = getAssets();
                 InputStream istr;
                 try {
                     istr = assetManager.open(gcmLargeIcon);
                     Bitmap bitmap = BitmapFactory.decodeStream(istr);
-                    mBuilder.setLargeIcon(bitmap);
+                    Bitmap bm = getCircleBitmap(bitmap);
+                    mBuilder.setLargeIcon(bm);
                     Log.d(LOG_TAG, "using assets large-icon from gcm");
```
macdonst:
```
if (roundIcon) {
  mBuilder.setLargeIcon(getCircleBitmap(bitmap);
} else {
  mBuilder.setLargeIcon(bitmap); 
}
```

bmwertman: Cool. I'm fine with making it an option. 

As far back as I've had Android devices and my memory serves, not sure which is longer, all large notifications have been round. The revision you have above is very similar to what I had at commit `9c351c2`. I was just having trouble getting the condition to catch that I had passed `roundIcon: true` to it. Would you want to see the condition set in `init` as @fr` suggested? Or do you like it as part of the ` data` payload?

Regarding copyright, I hadn't asked permission. Some experience with a former cofounder of mine led me into a discussion with a patent attorney. Basically only creative works are defensible as copyright. I wouldn't say something as standard as creating a circular image from a square one with an API falls within that realm. But I'm happy to post a question on the blog post and put a screenshot of the author's reply here if you like.

fredgalvao:
```
Large icons should be circular when showing a person, but square in all other cases.
```

https://material.io/guidelines/patterns/notifications.html#notifications-style

macdonst:
@fredgalvao yeah, I mentioned that in my comments as well.

macdonst:
@bmwertman you still working on this PR? I think it is a cool new feature.

bmwertman:
@macdonst I got a little side tracked in the holidays but I agree. I've got some things to get caught up on this week but I think I can get to it next week.

macdonst:
@bmwertman that's cool. I actually took time off during the holidays too. I was actually too hard to get to my laptop with all the family around. I'm just going over all the activity I missed and wanted to ping you to see if there is still interest. Since there is I'll leave it to you and merge when you are ready.

macdonst:
@bmwertman I cleaned up your code a bit and committed it as: https://github.com/phonegap/phonegap-plugin-push/commit/0dd0d468868f0b13c8d840c78dd89fca5920cd32

Basically I find the center of the image and draw a circle instead of an oval. It gets around the issue if the width and height are not equal. Otherwise you can get a shape that looks a lot like an egg. I also added a new parameter so you can tell the large icon to be displayed as a square (default) or circle.

Issue #1466, opened 2016-12-15
Duplicate notification from different SENDER_ID


TVolly:
My app has a plugin with the PushNotification service (for example Tapjoy). It receives pushes only from its own SENDER_ID.
Your plugin catches and displays all the pushes including other undesirable SENDER_IDs.
Thus, when those plugins (Tapjoy and others) send a push, it is being caught by two push services, and as a result I get 2 identical pushes .

Question: Is it possible to set a limit list of SENDER_ID so that your plugin could ignore it?
For example: add filter by "from" here:
https://github.com/phonegap/phonegap-plugin-push/blob/master/src/android/com/adobe/phonegap/push/GCMIntentService.java#L64

macdonst:
@TVolly yeah, it's possible but why do you have two services to receive push messages?

TVolly:
These plugins with different functionals. 
Push notification of that plugin is a bonus to its other features.
When a push sending from their service, it will be better if that notification would be catching by their plugin.
When a push sending from the server, that notification would be catching by your plugin

macdonst:
@TVolly uh sure. It's possible but it's not a high priority for me right now but pull requests are always welcome.

TVolly:
@macdonst I've sent a pull request with solution of this issue. We’re using your wonderful plugin as a dependency in our apps, so this issue is critical for us. Could you please review it? If there is something wrong we will fix it asap.

macdonst:
It will be part of 1.10.0.

PR #1475, opened 2016-12-20
Add sender ignore list for android

TVolly:
Issue: https://github.com/phonegap/phonegap-plugin-push/issues/1466


macdonst:
@TVolly wouldn't it be better just to have the plugin only use the sender ID that has been provided to it instead of a list of ID's to ignore?

TVolly:
Hello @macdonst 
I create new pull request: #1484

PR #1484, opened 2016-12-26
Receive notification only from SENDER_ID

TVolly:
Issue: #1466


macdonst:
Thanks for the PR @TVolly. Sorry it took so long to merge. It will be part of 1.10.0.

Issue #1140, opened 2016-08-03
Use VAPID on browser side


macdonst:
https://developers.google.com/web/updates/2016/07/web-push-interop-wins?hl=en

surajpindoria:
Thanks Simon, that article actually clears up a few things I was getting stuck on last week!

macdonst:
@surajpindoria np, I just figured it would make sense to `officially` assign it to you.

PR #1537, opened 2017-01-19
Added VAPID key support

heikoholland:
## Description
I added the possibility to pass a server key in the `init` method which is passed to the `pushManager.subscribe` method.

## Related Issue
#1140 

## Motivation and Context
I want to use the GCM push server for my browser app.

## How Has This Been Tested?
Tested in my app. If the server key is passed there is no need to post the subscribtion to the optinal given `pushServiceURL` so I won't execute that part of the code if an server key is given.

## Types of changes

## Checklist:

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -54,6 +54,7 @@ Attribute | Type | Default | Description
 Attribute | Type | Default | Description
 --------- | ---- | ------- | -----------
 `browser.pushServiceURL` | `string` | `http://push.api.phonegap.com/v1/push` | Optional. URL for the push server you want to use.
+`browser.applicationServerKey` | `string` | `` | Optional. Your GCM API key if you are using VAPID keys.
 
 #### iOS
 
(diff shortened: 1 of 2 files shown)


macdonst:
@heikoholland thanks for the PR. I'll have @surajpindoria look it over.

surajpindoria:
Hey @heikoholland,

In terms of code I think this is fine. Give me a chance to play around with it and I think this should be good to go!

surajpindoria:
Hi @heikoholland,

I am in the process of getting this all merged but had a quick question. Still trying to understand how the VAPID authentication works. In your code in `www/browser/push.js` on lines 75-76, you are not sending subscription information to the push server like we are currently doing in lines 66-73. Unless my understanding is incorrect, without doing this, when a push notification comes in the server won't have the information required to send it to the device.

Please correct me if I am wrong.

heikoholland:
You are right. I'm returning the complete data and send it to the server by myself in a via a JSON post request and couldn't use the form data way you are using.
Maybe we want to add one more option to specify the way we are sending the subscription information to the pushServiceURL?

macdonst:
@heikoholland can you check out the latest PR I just merged. It should take care of the VAPID issue. Regardless, I will close this PR.

Issue #1617, opened 2017-02-27
Handle payload differently in background or app-closed state


laagland:
Following up on #1596 

When a notification is received the code currently does not distinguish between a background state or closed-state. The notification payload is sent to the app and then either delivered or cached for delivery on next init.

Adding an extra check for this means the payload gets discarded when the app is not running. We can either add a **force-cache** or **no-cache** flag to the payload and handle it accordingly. Depending on preference of the default action, which one should we choose?

macdonst:
@laagland is this a comment on #1596 or a new issue? If it is a new issue please fill out the issue template.

laagland:
@macdonst this is a new issue, discussed in the comments of #1596. @fredgalvao advised to split the issues mentioned there. The issue template is basically the same. But to recollect:

### Expected Behaviour
According to the payload flowchart and the text beneath it, when the plugin receives the notification with content-available set to 1 and the app is not running (force-closed), no further processing should occur.

### Actual Behaviour
In the above situation, the payload is cached and sent to the app on initialisation.

While this could be intended behavior, with the current code there is no method to prevent the caching if you want to.

laagland:
@macdonst what should I do to move this issue forward. I've committed some changes to my fork that I think will handle this use-case.

macdonst:
@laagland well, I think you've done the best thing to move this forward by send a pull request. That is the best way to get anything done in an open source project. I'm stupid busy right now but hope to be able to review today and get this merged over the weekend.

laagland:
@macdonst awesome, thanks for your time!

macdonst:
This will be part of 1.10.0

PR #1620, opened 2017-03-02
Add no-cache flag to payload

laagland:
## Description
Add no-cache flag to payload, with default value '0'

## Related Issue
Issue #1617 

## Motivation and Context
According to the payload flowchart and the text beneath it, when the plugin receives the notification with content-available set to 1 and the app is not running (force-closed), no further processing should occur.

In the above situation, the payload is cached and sent to the app on initialisation.

While this could be intended behavior, with the current code there is no method to prevent the caching if you want to. This flag add a way to prevent it.

## How Has This Been Tested?
Only tested in local development environment

## Types of changes

## Checklist:


macdonst:
Thanks @laagland, I've merged it and it will be part of 1.10.0.

Issue #1663, opened 2017-04-03
Error on parsing GCM json structure


clementcontet:
### Expected Behaviour
When I send the following push notif to an Android:
```
{
    "message":
        {
            "title": "Title",
            "message" : "Message"
        }
}
```
I expect to get the correct title and message.

### Actual Behaviour
Yet, for the body of the notif, I get _{"title": "Title","message" : "Message"}_ instead of just Message

### Analysis
Digging in the code, the issue is that the plugin accept "message" as the json node for payload content (apparently since pull #218).
See:
```
      // If normalizeKeythe key is "data" or "message" and the value is a json object extract
            // This is to support parse.com and other services. Issue #147 and pull #218
            if (key.equals(PARSE_COM_DATA) || key.equals(MESSAGE) || key.equals(messageKey)) {
```

But later, the key for the body of the notif is normalized to the same string "message".
See:
```
    private String normalizeKey(String key, String messageKey, String titleKey) {
        if (key.equals(BODY) || key.equals(ALERT) || key.equals(MP_MESSAGE) || key.equals(GCM_NOTIFICATION_BODY) || key.equals(TWILIO_BODY) || key.equals(messageKey)) {
            return MESSAGE;
```

And finally, in `replaceKey` method, the body of my notif with the key "message" get replaced with the whole payload (with the same key "message").

Is there a point to this last `replaceKey` method call? (as all previously written key/values have already been normalized/localized)

#### Steps to Reproduce 
Send this payload to an Android

#### Platform and Version
Android 6.0.1

#### What device vendor
Nexus5

#### Cordova CLI version and cordova platform version

    cordova --version                                    6.5.0
    cordova platform version android                     6.1.2

#### Plugin version
    cordova plugin version | grep phonegap-plugin-push   1.10.0

#### Sample Push Data Payload
```
{
    "message":
        {
            "title": "Title",
            "message" : "Message"
        }
}
```

#### Logs taken while reproducing problem
04-03 17:02:16.016 20441-27607/? D/PushPlugin_GCMIntentService: onMessage - from: 520313199651
04-03 17:02:16.017 20441-27607/? D/PushPlugin_GCMIntentService: normalize extras
04-03 17:02:16.017 20441-27607/? D/PushPlugin_GCMIntentService: key = google.sent_time
04-03 17:02:16.017 20441-27607/? D/PushPlugin_GCMIntentService: replace key google.sent_time with google.sent_time
[...]

PR #1666, opened 2017-04-04
Fix issue #1663 by allowing "message" as the data payload key

clementcontet:
## Description
When receiving a message on Android, the normalizeExtras method tries to normalize each section in the following order:
1 If the key is holding the payload data structure
2 If the key is holding the notification data structure
3 For other keys

The problem was that 1/2 and 3 were not exclusive, the normalization process for "other key" (3) could happen just after the process for "payload data" (1). Which caused problems in case the same key "message" is used.

With this PR, the after the normalization process ran on the payload data, it won't try to normalize again the same key, erasing the previous value.

## Related Issue
Issue #1663

## Motivation and Context
It allows to have "message" as the payload data key. It is not standard, but it is used by some push services.

## How Has This Been Tested?
I tested the change on my device, with my "faulty" push service.
The changes are very limited so I'm very confident there won't be any side effects.

screenshot_20170404-093949

## Types of changes

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/GCMIntentService.java
+++ b/src/android/com/adobe/phonegap/push/GCMIntentService.java
@@ -253,11 +253,16 @@ public class GCMIntentService extends GcmListenerService implements PushConstant
                     newExtras.putString(newKey, valueData);
                 }
                 continue;
+            // In case we weren't working on the payload data node or the notification node,
+            // normalize the key.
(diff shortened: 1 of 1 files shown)


mjsalinger:
Is there an ETA on getting this merged?

macdonst:
@clementcontet thanks for the PR. Sorry it took so long to get merged. Been traveling around a lot for work these days.

PR #1672, opened 2017-04-06
fixed ios bug for gcm

LeftOfHere:
The code broke on ios 10.2 for us.
This fixed it.

Sorry for the fact that it is on master.

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -351,7 +351,7 @@
         //  GCM options
         [self setGcmSenderId: [iosOptions objectForKey:@"senderID"]];
         NSLog(@"GCM Sender ID %@", gcmSenderId);
-        if([[self gcmSenderId] length] > 0) {
+        if([self gcmSenderId]) {
             NSLog(@"Using GCM Notification");
             [self setUsesGCM: YES];
             [self initGCMRegistrationHandler];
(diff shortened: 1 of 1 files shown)


macdonst:
@jcesarmobile can you give this PR a look over?

jcesarmobile:
I don't have a GCM project to test.

But I don't see how this could fail if senderID is supposed to be a string in javascript and NSString on Objective-C, you shouldn't need the %ld formatting (long integer) 

@LeftOfHere are you configuring it like this?
	```
ios: {
                senderID: "12345679",
		alert: "true",
		badge: true,
		sound: 'false'
	}
```

Is this change is applied, the long integer formatting should also be done here https://github.com/LeftOfHere/phonegap-plugin-push/blob/fa9cb3728fdeb5e09ec51b97d7126013be92e9ce/src/ios/PushPlugin.m#L98

LeftOfHere:
I'll need to check tomorrow. Don't have the code with me. I also didn't write any of the app code, it was written only for android(with ios and windows phone as a known future endeavor) and I just needed to get it working on ios afterwards. 
Also just realized it's fcm not gcm, but doubt if that will make a difference, but anyway, will have a look tomorrow and let you know.

jcesarmobile:
I've tested and the app crash if I pass the senderID as integer instead of string, but it should be a string, so this fix is not needed.

BTW, if you want to use FCM, we have the v2.0.x branch in development for it

gianpaj:
yeah, this is wrong. it should be a string of integer (maybe Google will decide one day to add alpha chars?)

LeftOfHere:
The thing is, I add the plugin like the guide says,
cordova plugin add https://github.com/phonegap/phonegap-plugin-push --variable SENDER_ID="XXXXXXX"

And this gets generated in my plugin config.
"SENDER_ID": "XXXXXXX"

But it still thinks its a long in the obj C code.

jcesarmobile:
@LeftOfHere I mean your init method on javascript, you have to set the sender ID there and that's the value being read on `[self setGcmSenderId: [iosOptions objectForKey:@"senderID"]];`

LeftOfHere:
Thanks. We are using polymer and had a custom element setup to handle the config. The value was set to number in there.
Had it changed to a string and everything is working as expected now.

PR #1676, opened 2017-04-11
Fix the dates on the CHANGELOG.md (the last 2 version)

gianpaj:
all the previous dates need to be fixed as well. is a script broken somewhere? i couldn't find it


macdonst:
@gianpaj thanks for the PR. I merged it and I went and fixed the script that generates changelogs.

Issue #1665, opened 2017-04-04
App opens on clearing notification (Android)


dturakhia:
When I try to clear the notification from the notification centre by swiping it, instead of clearing the notification, it opens the app. I have checked my code and there is only one function that is called for handling notification with a condition to check if the app is closed or running in the background. 

```
    push.on('notification', function(data) {
        if (data.additionalData.coldstart == true){
                 //Do Something
        } else {
                 //Do Something                
        }
```

Is it a bug in the code or am I doing something wrong? I have tested this on multiple devices with multiple OS. The devices tested are -

One Plus 3T (Android 7.0)
Nexus 5 (Android 6.0.1)
Samsung Note 3 Neo (Android 5.1.1)

All these devices face the same issue.

macdonst:
@dturakhia that should not happen at all. What version of the plugin are you using?

dturakhia:
I'm currently using version **1.10.0** for this project. My earlier projects did not face any such issue where I was using 1.9.0

youngsun-kwon:
I have the same problem. 
Android 7.0, 6.0.1

remcoanker:
I had the same problem. I fixed it by removing the `.setDeleteIntent(deleteIntent)` in the notification builder.

[this line](https://github.com/phonegap/phonegap-plugin-push/blob/d669c44a863d86d0bb73b5ae086bc2fe6f8113a9/src/android/com/adobe/phonegap/push/GCMIntentService.java#L354)

macdonst:
Well, crap. I can reproduce this. Let me get to fixing it immediately.

macdonst:
@dturakhia I released 1.10.1 to work around this and I'm going to see if @laagland can do a real fix. 

@laagland can you see if when you dismiss the notification it does not start the app?

laagland:
Yes, I'll have a look

dturakhia:
@macdonst Thanks mate, the 1.10.1 update seems to have the fixed the issue. Shall wait for the real fix, until then this should work fine.

macdonst:
@laagland did you have any luck finding a way to dismiss the notification without starting the app?

PR #1816, opened 2017-06-30
V1.x Android: handle dismissed notifications

NitroGhost:
## Description

When a user dismisses notifications, you would expect them to go away. Currently when using Inbox style, previous notifications will be included, even though the user has dismissed the notification. This fix handles dismissed notifications, and clears the internal list.
## Related Issue

#1665

## Motivation and Context

With the related issue, originally the dismissal of the notification would result in the app opening which is obviously the incorrect way to handle this, but the "fix" applied was just to remove the deleteIntent altogether, which is also the incorrect way to handle this.
## How Has This Been Tested?

I sent a bunch of inbox style notifications with the same notId, dismissed the notification, then resent a bunch more. Checked to see if the previous dismissed notifications were gone. They were.
## Types of changes

## Checklist:

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -40,6 +40,7 @@
     <config-file target="AndroidManifest.xml" parent="/manifest/application">
       <activity android:name="com.adobe.phonegap.push.PushHandlerActivity" android:exported="true" android:permission="${applicationId}.permission.PushHandlerActivity"/>
       <receiver android:name="com.adobe.phonegap.push.BackgroundActionButtonHandler"/>
+      <receiver android:name="com.adobe.phonegap.push.PushDismissedHandler"/>
       <receiver android:name="com.google.android.gms.gcm.GcmReceiver" android:exported="true" android:permission="com.google.android.c2dm.permission.SEND">
         <intent-filter>
           <action android:name="com.google.android.c2dm.intent.RECEIVE"/>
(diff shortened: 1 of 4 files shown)


fredgalvao:
Doesn't this go against some common cases of conversation/threaded applications? Such as:

- Maria sends you "Hey"
- Maria sends you "What's up"
- You dismiss the single inbox-style notification that had those two messages
- Maria sends you "Call me back"
- ?

At this point, most applications would bring back the 2 you dismissed (as they actually weren't processed, just dismissed/delayed), agregate the third one, and present the user a new inbox-style notification with all 3 of them.

The interpretation of what `dismissing a notification` means is subjective, I know, and multiple scenarios would argue to both sides, and thus I'd say we have to at least let the user decide instead of assume.

fredgalvao:
Also, we might need to bring whatever is decided here to the master branch too.

macdonst:
@fredgalvao this PR deals with Inbox style notifications not Android N groupings. When you dismiss an inbox style notification all of the notifications should be dismissed from the plugin as they are only contained in one notification. In Andoid N they can be expanded to have multiple lines in the notification center. 

This also fixes the `setDeleteIntent` problem. This is a good basis for either fire the `notification` event with `dismissed = true` or adding a new `dismissed` event.

fredgalvao:
I should've read the related issue at full before commenting. Nevermind my daydreams.

NitroGhost:
Would you like me to do a v2.x pull request also for this?

macdonst:
@NitroGhost nope, I was doing a cherry-pick and my computer crashed. Should be committed soon.

NitroGhost:
that's great 😁

macdonst:
@NitroGhost harsh! You are happy my computer crashed 😉

macdonst:
@NitroGhost merged into v2.x

Issue #1536, opened 2017-01-19
Typescript definitions are outdated, API needs remirroring


fredgalvao:
### Expected Behaviour
100% API and settings/options coverage on typescript

### Actual Behaviour
We're lacking `clearAllNotifications` there, at the very least. Needs reviewing.

#### Steps to Reproduce 
See #1530.

#### Plugin version
Latest on each main branch (1.9.x && 2.0.x)

fredgalvao:
Update #1621 in type definition too.

macdonst:
@fredgalvao can we close this now?

fredgalvao:
Nop. Having the definitions in-house made it easier, but not invalid. I'll still work on updating the interfaces/objects.

kanak-infoObjects:
I am facing the same issue. Can't find the clearAllNotifications function. Please suggest a solution or an alternative. Thank you.

fredgalvao:
@kanak-infoObjects  I don't have an alternative. I'm sorry to say I haven't had the time to review the type definitions since forever, I'll try to get it done soon.
As always PRs and contributions are welcome.

fredgalvao:
Even though `clearAllNotifications` is now typed, the full API still needs reviewing.

fredgalvao:
I'll close this and open a new one to actually update all the missing parts, as this is too old by now, and has been partially done over time.

fredgalvao:
It's time.

PR #1827, opened 2017-07-04
Add clearAllNotifications() to type definitions

szh:
Adds the `clearAllNotifications()` function to the type definitions

## Related Issue
Fixes https://github.com/phonegap/phonegap-plugin-push/issues/1536

## Motivation and Context
Updates type definitions to match API changes

## Types of changes

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -103,6 +103,13 @@ declare namespace PhonegapPluginPush {
 		 * @param id
 		 */
 		finish(successHandler?: () => any, errorHandler?: () => any, id?: string): void
+		
+		/**
+		 * Tells the OS to clear all notifications from the Notification Center
+		 * @param successHandler Is called when the api successfully clears the notifications.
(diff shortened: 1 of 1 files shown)


fredgalvao:
@macdonst While reviewing the android source code for this, I found that the `clearAllNotifications` action does not handle any type of failure explicitly:
```java
else if (CLEAR_ALL_NOTIFICATIONS.equals(action)) {
            cordova.getThreadPool().execute(new Runnable() {
                public void run() {
                    Log.v(LOG_TAG, "clearAllNotifications");
                    clearAllNotifications();
                    callbackContext.success();
                }
            });
        }
```

Would that entail that it will never fail? Should we try/catch it for basic failure handling? Is there any argument to any of those callbacks in any case?

macdonst:
@fredgalvao if this addition to the types file looks good please go ahead and merge it.

WRT clearAllNotifications in Android, I'm not even sure if it can fail but create and issue for it and I'll do a quick look.

fredgalvao:
@szh Thanks!

PR #1934, opened 2017-09-01
Return error message if user tapped "Don't Allow"

extempl:
Based on https://stackoverflow.com/a/34189489/506954

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -291,8 +291,8 @@
                                                       //user tapped "Allow"
                                                   }
                                                   else {
-                                                      NSLog(@"User didn't alowed Notifications for the App");
-                                                      [self failWithMessage:self.callbackId withMsg:@"User didn't alowed Notifications for the App" withError:Nil];
+                                                      NSLog(@"User didn't allowed Notifications for the App");
+                                                      [self failWithMessage:self.callbackId withMsg:@"User didn't allowed Notifications for the App" withError:Nil];
                                                   }
                                               }];
         }
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -282,6 +282,19 @@
             UIUserNotificationSettings *settings = [UIUserNotificationSettings settingsForTypes:UserNotificationTypes categories:categories];
             [[UIApplication sharedApplication] registerUserNotificationSettings:settings];
             [[UIApplication sharedApplication] registerForRemoteNotifications];
+          
+            [[NSNotificationCenter defaultCenter] addObserverForName:UIApplicationDidBecomeActiveNotification
+                                                  object:nil
+                                                   queue:[NSOperationQueue mainQueue]
+                                              usingBlock:^(NSNotification * _Nonnull note) {
+                                                  if ([[UIApplication sharedApplication] isRegisteredForRemoteNotifications]) {
+                                                      //user tapped "Allow"
+                                                  }
+                                                  else {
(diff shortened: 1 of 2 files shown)


macdonst:
@extempl I've been testing this but I can't get it to fail when the user clicks "Don't Allow" the code never seems to get called. Can you give me a better reproduction scenario?

extempl:
@macdonst That was a bit obsolete version, I've updated it to how it works on my end now:
Added two events `allowed` and `disallowed`. For checking access it uses same approach as used in `hasPermission` method. 
Also added ability to `off` all handlers of certain event, as `allowed` will fire each time app is resumed, so, on my end, I am listening for `allowed`/`disallowed` event only once before the first register. For `resume` I using built-in `hasPermissions` method.

Probably the code is need in polishing, I am not an ios-dev.

src/ios/PushPlugin.m
```diff
@@ -282,6 +282,28 @@ - (void)init:(CDVInvokedUrlCommand*)command;
...
+                                                                                                    queue:[NSOperationQueue mainQueue]
+                                                                                               usingBlock:^(NSNotification * _Nonnull note) {
+               [NSThread sleepForTimeInterval:0.1f];
+               BOOL enabled = NO;
+               id<UIApplicationDelegate> appDelegate = [UIApplication sharedApplication].delegate;
+               if ([appDelegate respondsToSelector:@selector(userHasRemoteNotificationsEnabled)]) {
+                   enabled = [appDelegate performSelector:@selector(userHasRemoteNotificationsEnabled)];
+               }
+               if (enabled) {
+                   NSLog(@"User allowed Notifications for the App");
```
macdonst: I'm cool with the log but I think we can lose everything else in this if.

www/push.js
```diff
@@ -34,7 +34,9 @@ var PushNotification = function () {
     this.handlers = {
       registration: [],
       notification: [],
-      error: []
+      error: [],
+      allowed: [],
```
macdonst: Let's get rid of the allowed/disallowed event listeners.

www/push.js
```diff
@@ -47,7 +49,9 @@ var PushNotification = function () {
 
     // triggered on registration and notification
     var success = function success(result) {
-      if (result && typeof result.registrationId !== 'undefined') {
+      if (result === 'allowed') {
```
macdonst: Scratch this.

src/ios/PushPlugin.m
```diff
@@ -282,6 +282,28 @@ - (void)init:(CDVInvokedUrlCommand*)command;
...
+                   CDVPluginResult* pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK messageAsString:@"allowed"];
+                   [pluginResult setKeepCallbackAsBool:YES];
+                   [self.commandDelegate sendPluginResult:pluginResult callbackId:self.callbackId];
+               }
+               else {
+                   NSLog(@"User disallowed Notifications for the App");
+                   [self failWithMessage:self.callbackId withMsg:@"disallowed" withError:Nil];
```
macdonst: Maybe just give a more detailed error message here.

www/push.js
```diff
@@ -77,6 +81,9 @@ var PushNotification = function () {
 
     // triggered on error
     var fail = function fail(msg) {
+      if (msg === 'disallowed') {
```
macdonst: Probably just get rid of this and let the regular error handler take care of it.

PR #1945, opened 2017-09-08
Update INSTALLATION.md

dnldsht:
fixed issue google-services.json not found


macdonst:
@dnldsht does this PR have an issue associated with it? I'm not sure what problem it is supposed to solve. When my config.xml omits the `/` the file gets delivered properly.

Issue #1973, opened 2017-09-22
Unsubscribe from topic duplicated action


Chuckytuh:
### Expected Behaviour
Only one call to `FirebaseMessaging.getInstance().unsubscribeFromTopic(topic);`
### Actual Behaviour

`FirebaseMessaging.getInstance().unsubscribeFromTopic(topic);` is being called twice when unsubscribing a topic.

macdonst:
Currently fixed in master, soon to be in release 2.1.0.

PR #1974, opened 2017-09-22
Remove duplicated action to unsubscribe topic

Chuckytuh:
## Description

There's no need to call the unsubscribe twice.

## Related Issue

PR for issue #1973


macdonst:
Good catch @Chuckytuh, thanks!

Issue #2130, opened 2018-01-04
Delete the firebase instance when unregistering the firebase token


ccorcos:
### Expected Behaviour

When I call `Push.unregister(() => {}, () => {})`, I would expect this plugin to call `[[FIRInstanceID instanceID] deleteIDWithHandler]` to actually unregister the application token with Firebase ([example](https://github.com/arnesson/cordova-plugin-firebase/blob/master/src/ios/FirebasePlugin.m#L180)). That way, when I try to send a message to this token, firebase will respond with `messaging/registration-token-not-registered`. 

### Actual Behaviour

We simply [disable the ability to receive notifications](https://github.com/phonegap/phonegap-plugin-push/blob/master/src/ios/PushPlugin.m#L122) but Firebase will respond as if the message succeeded.

```
"cordova": "7.1.0",
<engine name="ios" spec="4.5.4" />
"name": "phonegap-plugin-push",   "version": "2.1.2",
```

macdonst:
@ccorcos please update your issue to match the issue template.

Issue #2082, opened 2017-11-29
How to force refresh GCM token (registration ID) in iOS?


malwatte:
I tried unregistering and re-registering to force GCM token refresh. But it did not refresh the GCM token. (I use GCM to send push notifications.)

1. I registered (init the plugin) and sent push notifications. App received notifications.
2. I unregistered and sent push notifications to the app. App did not received push notifications.
    That is expected.
    But the problem is I thought I would get get "NotRegistered" error from FCM as the response 
    when calling FCM end-point. Instead it returned a messageId. That means notification is send 
    from FCM but OS did not showed it in the device.
3. I waited for 15 minuted and sent notifications. But still same behavior.
4. I registered (init the plugin) again and plugin returned the old GCM token. I though I would get a 
    new token. I sent notifications and app received notifications. That means token is not 
    refreshed.

Problem is in iOS. It works fine in Android. Android returned a new token.

Plugin version: 1.8.4 (We are still using ionic 1. Cannot migrate to newer version right now)
iOS: tested in 8.4 and 11.0.3

I checked the code and found that
Android: 
When unregistering, token is deleted by `InstanceID.getInstance(getApplicationContext()).deleteInstanceID();`
but iOS:
` [[UIApplication sharedApplication] unregisterForRemoteNotifications];`

I Googled and checked how I can delete and refresh the token in iOS.
I found that I need to do this to delete and get a new token in iOS.

```
GGLInstanceID *iid = [GGLInstanceID sharedInstance];
GGLInstanceIDDeleteHandler deleteHandler = ^void(NSError *error) {
  if (error) {
    // failed to delete the identity for the app
    // do an exponential backoff and retry again.
  } else {
    // try to get a new ID
    dispatch_async(dispatch_get_main_queue(), ^{
       GGLInstanceIDHandler handler =
           ^void(NSString *identity, NSError *error) {
              if (error) {
                // failed to get the identity for the app
                // handle error
              } else {
                NSString *instanceID = identity;
                // handle InstanceID for the app
              }
           }
       [iid getIDWithHandler:handler];
    });
  }
}

[iid deleteIDWithHandler:deleteHandler];
```
(Refer https://developers.google.com/instance-id/guides/ios-implementation)

Question: How I can force a token refresh in iOS?

malwatte:
I ended up in modifying the plugin version 1.8.4. I added some code to delete the InstanceID.
[...]

PR #2131, opened 2018-01-04
Unregister token with firebase

ccorcos:
Solves https://github.com/phonegap/phonegap-plugin-push/issues/2130

## Description
Rather than disable the ability to receive notifications locally on the phone, we're actually unregistering the token with FCM.

## Motivation and Context
A user logs in, we save the token for that user, the user logs out so we call unregister. A new user logs in. These two users should not be sharing the same push notification token.

## How Has This Been Tested?
I used this code with my app. Logged in as a user, was able to receive notifications, called unregister on logout, attempted to send a message to that same token and received a `messaging/registration-token-not-registered` error message on the server and never received the message on the phone.

## Types of changes
This isn't really a breaking change for the plugin, but it is logically different in terms of application code that uses this plugin. They may start to see these unregistered errors. They also can't expect to unregister and then re-init and keep the same token. But I don't imagine these are reasonable expectations.

## Checklist:

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -119,8 +119,9 @@
             [pubSub unsubscribeFromTopic:topic];
         }
     } else {
-        [[UIApplication sharedApplication] unregisterForRemoteNotifications];
-        [self successWithMessage:command.callbackId withMsg:@"unregistered"];
+        [[FIRInstanceID instanceID] deleteIDWithHandler:^void(NSError *_Nullable error){
+            [self successWithMessage:command.callbackId withMsg:@"unregistered"];
+        }];
     }
 }
 


src/ios/PushPlugin.m
```diff
@@ -119,8 +119,9 @@ - (void)unregister:(CDVInvokedUrlCommand*)command;
             [pubSub unsubscribeFromTopic:topic];
         }
     } else {
-        [[UIApplication sharedApplication] unregisterForRemoteNotifications];
```
macdonst: @ccorcos you can't delete this section as it is required for users who are using APNS to send messages. You have found an issue though and the code should check to see if `usesFCM` is true. If it is then you should call the `deleteIDWithHandler` method.

macdonst:
@macdonst a reminder to also look at Issue #2082 and this commit https://github.com/delphist2008/phonegap-plugin-push/commit/365e1a5336cc8a10a5eb2efc5e098c81d7c9f912

macdonst:
Superceeded by #2509

Issue #2191, opened 2018-02-06
Clear One Notification (Android)


DavidBriglio:
### Expected Behaviour
To have a method that will clear one single notification, on the Android platform.

### Actual Behaviour
There only exists a `clearAllNotifications` method that will remove all notifications.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Android (all)

**None of the other fields apply to this.**

---

I would like to make a PR that will add single notification removal support for the android platform. This is specifically in regard to clearing ongoing notifications. If there are multiple notifications being displayed, and you want to remove an ongoing notification your only option is to remove all notifications.

macdonst:
Android and iOS support in 3.0.0.

PR #2192, opened 2018-02-06
:sparkles: Added a clearNotification method to clear a single notification (Android Only)

DavidBriglio:
## Description
This PR contains an additional method named `clearNotification` that will clear one single notification based on the ID argument. This also uses a success and error callback on completion.

## Related Issue
Issue #2191

## Motivation and Context
As mentioned in the issue, I would like a method to specifically remove an ongoing notification. This will allow for any notification to be removed based on the notification id number.

## How Has This Been Tested?
I have added tests to the test suite, and have tested on (cordova@7.1.0):
- Pixel running Android 8.1 (cordova-android@6.4.0)
- iPhone 6S running iOS 11.2.5 (cordova-ios@4.5.2)

## Types of changes

## Checklist:
- [X] I have added tests to cover my changes.


docs/API.md
```diff
@@ -506,3 +507,25 @@ push.clearAllNotifications(() => {
...
+
+Tells the OS to clear the notification that corresponds to the id argument, from the Notification Center
+
+### Parameters
+
+Parameter | Type | Default | Description
+--------- | ---- | ------- | -----------
+`id` | `number` | | The ID of the notification that will be cleared. |
+`successHandler` | `Function` | | Is called when the api successfully clears the notifications.
+`errorHandler` | `Function` | | Is called when the api encounters an error when attempting to clears the notifications.
```
fredgalvao: [typo] `notifications` here should not be plural. There's a typo on `attempting to clears` too, but that is also wrong on the original `clearAllNotifications` doc, so we could fix them both.

docs/API.md
```diff
@@ -506,3 +507,25 @@ push.clearAllNotifications(() => {
...
+
+### Example
+
+```javascript
+push.clearNotification(145, () => {
```
fredgalvao: I wonder if it's worth to put the `id` argument last, to match the rest of the API. Although I think we got the order wrong on the rest, not here, but then to break it all would be a huge pain...

src/js/push.js
```diff
@@ -203,6 +203,18 @@ class PushNotification {
 
     exec(successCallback, errorCallback, 'PushNotification', 'clearAllNotifications', []);
   }
+
+  /**
+   * Clears notifications that have the ID specified.
+   * @param  {[type]} id    ID of the notification to be removed.
+   */
+  clearNotification(id, successCallback = () => {}, errorCallback = () => {}) {
+    if (id && !isNaN(parseInt(id, 10))) {
```
fredgalvao: Considering the backend needs the `notId` to be an `int`, and even needs it to not go beyond `MAX_VALUE`, shouldn't we yield an error if the `id` is prone to error?

fredgalvao:
Hey, thanks for contributing! I left some comments regarding what I could find quickly, please let me know what you think of them.

docs/API.md
DavidBriglio: I will make these changes in the next commit.

docs/API.md
DavidBriglio: I think that consistency is probably best here. Since the other methods are using `id` last, I will change this method to use it last as well. This way if in the future if the API is changed to use `id` first, this method will just be changed with the rest.

src/js/push.js
DavidBriglio: Please look at my updated code, I have added more error checking (can't be below 0 or above MAX_SAFE_INTEGER) and added an error message. I don't believe there is any more error checking required on the back end, since the try catch will catch an error while parsing the integer out. Please let me know if these changes are sufficient.

macdonst:
@DavidBriglio this is a solid contribution but I've been hesitant to introduce any API call into this plugin that only works on one platform. If this was to work on Android/iOS then it would be more compelling. It's been awhile since I've looked into the possibility of this working on iOS. Have you checked?

DavidBriglio:
It looks like there may be a way to do this in iOS. I will do more testing and see if I can add it to this PR.

fredgalvao:
@DavidBriglio  All the comments were fixed, the validations are much better now, thanks!

DavidBriglio:
After some trial and error I was not able to implement this myself on iOS, but I think that these methods could be used to remove notifications:
- [removeDeliveredNotificationsWithIdentifiers](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649500-removedeliverednotificationswith?language=objc)
- [removePendingNotificationRequestsWithIdentifiers](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649517-removependingnotificationrequest?language=objc)
- [getPendingNotificationRequestsWithCompletionHandler](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649513-getpendingnotificationrequestswi?language=objc)
- [getDeliveredNotificationsWithCompletionHandler](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649520-getdeliverednotificationswithcom?language=objc)

macdonst:
Cool, I'm going to get back to this soonish. We should be able to update our iOS implementation and then add this in as a new feature for both platforms.

merrygobyebye:
@macdonst, any assistance needed with the iOS implementation of this feature?

macdonst:
@DavidBriglio thanks for the PR, it is finally merged and I'll be doing a 3.0.0 release with it soon.

@merrygobyebye sure, go for it. With the update to UNUserNotification this is now possible on iOS.

merrygobyebye:
@macdonst Sure -- I'll try to get it done within the next several days.

Issue #2248, opened 2018-03-06
Android Notifications fail if other channels are already setup


dpeacock:
### Expected Behaviour
The plug-in should allow for other plug-ins already having channels.  It should create a new channel if it doesn't yet have one.

### Actual Behaviour
It only checks to see if any channels exist, not our channel explicitly.  This causes us not to create a channel and the notifications not to display.

### Reproduce Scenario (including but not limited to)
You will need another plug-in that is using notification channels.  If this plug-in initializes first then we will fail to create our channel in createDefaultNotificationChannelIfNeeded as we only check to see if the size is 0.

#### Steps to Reproduce 
See above.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Occured in Android 8.0

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
OnePlus 5

#### Cordova CLI version and cordova platform version
    cordova --version                                    8.0.0
    cordova platform version android                     6.4.0
#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   2.1.3

macdonst:
@dpeacock this will be included in the next release.

PR #2249, opened 2018-03-06
Android fix to check if the channel is setup.

dpeacock:
## Description
Previously we only created our channel if no channels were created, there could be other channels created already so this would leave us without our channel.  We should explicitly check if our channel exists, and not assume we are the only channel.

## Related Issue
https://github.com/phonegap/phonegap-plugin-push/issues/2248

## Motivation and Context
If you are using another plugin that uses notification channels the plug-in will not display the notifications.

## How Has This Been Tested?
Tested in our production application where notifications were failing when we had other plugins using notification testing.  Tested on OnePlus 5 running Android 8.0.
## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


macdonst:
@dpeacock thanks for the PR!

PR #2293, opened 2018-03-30
Add dependency on cordova-support-google-services

chemerisuk:
Adding dependency on [cordova-support-google-services](https://github.com/chemerisuk/cordova-support-google-services) should resolve some conflicts.

## Description
Per our discussion at https://github.com/chemerisuk/cordova-support-google-services/issues/1#issuecomment-377402542. 

## Related Issue
https://github.com/chemerisuk/cordova-support-google-services/issues/1

## Motivation and Context
Solve compatibility issues between cordova plugins.

## Types of changes

## Checklist:

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -44,6 +44,7 @@
     <framework src="me.leolin:ShortcutBadger:1.1.17@aar"/>
     <framework src="com.google.firebase:firebase-messaging:$FCM_VERSION"/>
     <framework src="push.gradle" custom="true" type="gradleReference"/>
+    <dependency id="cordova-support-google-services" version="~1.1.0"/>
     <source-file src="src/android/com/adobe/phonegap/push/FCMService.java" target-dir="src/com/adobe/phonegap/push/"/>
     <source-file src="src/android/com/adobe/phonegap/push/PushConstants.java" target-dir="src/com/adobe/phonegap/push/"/>
     <source-file src="src/android/com/adobe/phonegap/push/PushHandlerActivity.java" target-dir="src/com/adobe/phonegap/push/"/>
--- a/push.gradle
+++ b/push.gradle
@@ -17,20 +17,3 @@ android {
         applicationId = doExtractStringFromManifest("package")
     }
 }
-
-buildscript {
-	repositories {
-      jcenter()
-			mavenLocal()
-  }
-  dependencies {
-      classpath 'com.android.tools.build:gradle:+'
-      classpath 'com.google.gms:google-services:3.0.0'
-  }
-}
-
-// apply plugin: 'com.google.gms.google-services'
-// class must be used instead of id(string) to be able to apply plugin from non-root gradle file
-ext.postBuildExtras = {
-    apply plugin: com.google.gms.googleservices.GoogleServicesPlugin
-}


chemerisuk:
@macdonst let me know if I can help to update documentation.

macdonst:
@chemerisuk thanks for the PR. I just merged it in. We should try and get as many people as possible to move to using your plugin as a dependency. Actually, we should probably make it an option you can enabled in `cordova-android`.

chemerisuk:
@macdonst agree, It's hard to imagine modern Android development without having a dependency on google services. The advantage of gradle plugins is that they do not add extra code at runtime, so people that for a some reason do not have `google-services.json` won't notice any difference.

The only issue is compatibility with many existing plugins... On the other hand recent `cordova-android 7` breaks plugins anyway, so may be it's a good to to integrate `cordova-support-google-services` into the core. Thoughts?

fredgalvao:
@chemerisuk You should try and spread that word on the cordova dev mailing list. We'll get broader opinions there.

macdonst:
I agree with @fredgalvao, we should bring this to the cordova dev mailing list. What I envision is a preference like:

```
<preference name="android-RunGoogleServicesPlugin" value="true"/>
```

where the default it is `false`. Then any app that needs the plugin to run will enable this preference. 

Say this lands in cordova-android 8 then @chemerisuk can update the `cordova-support-google-services` so it only installs in an app when cordova-android is less than 8. That way plugin devs who use the `cordova-support-google-services` won't even need to update their plugin.xml files.

Also, @chemerisuk what do you mean by cordova-android 7 breaks plugins?

chemerisuk:
@macdonst folder structure has changed and in many cases plugins must be updated because of that (read http://cordova.apache.org/announcements/2017/12/04/cordova-android-7.0.0.html#disqus_thread).

As for preference `android-RunGoogleServicesPlugin` - it totally makes sense. I was thinking first about always adding the gradle plugin (because it doesn't add any runtime code), but after testing on empty project it seems like the gradle plugin fails without `google-services.json`:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:processDebugGoogleServices'.
> File google-services.json is missing. The Google Services Plugin cannot function without it. 
   Searched Location: 
  /.../git/test/platforms/android/app/src/nullnull/debug/google-services.json
  /.../git/test/platforms/android/app/src/debug/nullnull/google-services.json
  /.../git/test/platforms/android/app/src/nullnull/google-services.json
  /.../git/test/platforms/android/app/src/debug/google-services.json
  /.../git/test/platforms/android/app/src/nullnullDebug/google-services.json
  /.../git/test/platforms/android/app/google-services.json
```

macdonst:
@chemerisuk I just sent a message out to the dev mailing list. Please chime in.

chemerisuk:
@macdonst I have issue with connecting to that discussion... The only suggestion is to drop `android-` prefix from the preference name - it seems like new cordova preferences use only case case style, may be better to be consistent. Something like `RunGoogleServicesPlugin` or `EnableGoogleServicesPlugin`.

macdonst:
@chemerisuk sorry are you saying you can't participate in the discussion? With regards to the exact name of the preference I'm not worried about what it eventually gets called.

chemerisuk:
@macdonst had problems with authentication, now resolved.

chemerisuk:
@macdonst  Just submitted PR at https://github.com/apache/cordova-android/pull/438

Issue #2305, opened 2018-04-06
Improve Docs about Payload: Add description of sample JSON message format


ptandler:
### Expected Behaviour

I'm quite new to cordova and push notifications and it took me quite some time to figure out how the message / payload formats should be for the different platforms. So I started adding some sample json to the PAYLOAD.md as I think this is the place I would have looked for this.

I started writing in my fork, see https://github.com/teambits/phonegap-plugin-push/blob/master/docs/PAYLOAD.md#message-format-overview

#### Cordova CLI version and cordova platform version

    cordova --version                                    # e.g. 6.0.0
    cordova platform version android                     # e.g. 4.1.1

#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   # e.g. 1.5.3

PR #2306, opened 2018-04-06
Added description of sample JSON message format for Android and iOS to PAYLOAD.md

ptandler:
## Description

I'm quite new to cordova and push notifications and it took me quite some time to figure out how the message / payload formats should be for the different platforms. So I started adding some sample json to the PAYLOAD.md as I think this is the place I would have looked for this.

I started writing in my fork, see https://github.com/teambits/phonegap-plugin-push/blob/master/docs/PAYLOAD.md#message-format-overview

## Related Issue

https://github.com/phonegap/phonegap-plugin-push/issues/2305

## Motivation and Context

Well, I think it's easier to get started using PN :smile: 

## How Has This Been Tested?

Well, I pasted some of the sample JSON to the SNS web UI to see if it's really working.
But I didn't do any other kind of test yet such as having someone else read all this.

## Types of changes

## Checklist:

### DIFF
--- a/docs/PAYLOAD.md
+++ b/docs/PAYLOAD.md
@@ -16,7 +16,7 @@
   - [Priority in Notifications](#priority-in-notifications)
   - [Picture Messages](#picture-messages)
   - [Background Notifications](#background-notifications)
-    - [Use of content_available: true](#use-of-content-available-true)
+    - [Use of content-available: true](#use-of-content-available-true)
   - [Caching](#caching)
   - [Huawei and Xiaomi Phones](#huawei-and-xiaomi-phones)
   - [Application force closed](#application-force-closed)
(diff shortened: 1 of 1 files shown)


docs/PAYLOAD.md
```diff
@@ -72,6 +72,115 @@ Some ways to handle this *double* event are:
...
+
+## iOS Message Format
+
+The JSON message can contain the following fields, see [Apple developer docs](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html#//apple_ref/doc/uid/TP40008194-CH17-SW5) for a complete list
+
+```javascript
+{
+    "aps":{
+        "alert": { // alternatively just a string: "Your Message",
+			"title": "A short string describing the purpose of the notification",
```
fredgalvao: The blocks above are properly formatted, but the ones bellow are not. I think we could use a proper json formatting through all examples/snippets.

fredgalvao:
Thanks for this! We've been needing such a section for a while now.

docs/PAYLOAD.md
```diff
@@ -16,7 +16,7 @@
   - [Priority in Notifications](#priority-in-notifications)
   - [Picture Messages](#picture-messages)
   - [Background Notifications](#background-notifications)
-    - [Use of content_available: true](#use-of-content-available-true)
+    - [Use of content-available: true](#use-of-content-available-true)
```
fredgalvao: The wrong part of this bad link is the ID, not the name. This should instead be fixed to:
`- [Use of content_available: true](#use-of-content_available-true)`

fredgalvao:
I also think we could use a section dedicated to FCM on iOS, as the data structure is certainly at odds with the APNS one.

docs/PAYLOAD.md
ptandler: ok, fixed

docs/PAYLOAD.md
ptandler: You mean this one:

`{"GCM":"{\"priority\":\"normal\",\"content_available\":\"0\",\"data\":{\"title\":\"A short string describing the purpose of the notification\",\"message\":\"The text of the alert message\",\"count\":5,\"notId\":1,\"custom_key1\":\"value1\",\"custom_key2\":\"value2\"}}"}`?

The problem here is that this is a very long string and AFAI it cannot be split into several lines in JSON. I could write

`{
"GCM":
"{\"priority\":\"normal\",\"content_available\":\"0\",\"data\":{\"title\":\"A short string describing the purpose of the notification\",\"message\":\"The text of the alert message\",\"count\":5,\"notId\":1,\"custom_key1\":\"value1\",\"custom_key2\":\"value2\"}}"
}`

but I fear this doesn't help much ... what do you think?

ptandler:
@fredgalvao WRT FCM, I have no experience with this myself (yet), so I can't contribute here.

docs/PAYLOAD.md
fredgalvao: I was actually refering to the "normal" payload sections. There were a few parts where an inner object was indented with tabs instead of spaces, and even though I don't have a problem with tabs per se, they have mixed whitespace usage. We should stick to one of them for consistency.

The "generated" payload examples you provided need not be formatted, I agree on that.

ptandler: alright, I'll fix the whitespaces, didn't notice.

ptandler: I just checked in a version with corrected whitespace and "sound" attribute added. What do you think now?

docs/PAYLOAD.md
```diff
@@ -16,7 +16,7 @@
   - [Priority in Notifications](#priority-in-notifications)
   - [Picture Messages](#picture-messages)
   - [Background Notifications](#background-notifications)
-    - [Use of content_available: true](#use-of-content-available-true)
+    - Use of content_available: true](#use-of-content_available-true
```
macdonst: @ptandler I believe you accidentally removed the open and closing brackets from this line which will cause the href to be broken

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
 - send two pushes, one to be processed in the background, and the other to show up in the shade.
 - include a unique ID in your push so you can check to see if you've already processed this event.
 
+# Message Format Overview
```
macdonst: This is missing from the table of contents.

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
...
+
+## Android Message Format
+
+The JSON push message can contain the following fields, see https://developers.google.com/cloud-messaging/http-server-ref for a complete list.
+
+```javascript
+var content = {
+  "priority": "normal", // Valid values are "normal" and "high."
+  "content_available": "0", // see "Background Notifications" below
```
macdonst: No, this is not the correct location for this property for this plugin. Yes, I know this is where the above reference says it should be but it will screw up the receipt of the incoming push.

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
...
+  "data": {
+    "title": "A short string describing the purpose of the notification",
+    "message": "The text of the alert message", // "body" can be used as alias, is converted to "message"
+    // localozation of message is possible
```
macdonst: Should be `localization` not `localozation`

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
...
+    "count": 5, // set the badge notification count at app icon
+    "sound": "default", // play default sound ... or "soundname", see [Android Sound](#sound) section
+    "notId": 1, // unique ID for the message, used for grouping, see below
+    "custom_key1": "value1",
+    "custom_key2": "value2"
+  },
+  // "notification": { ... } // can be used instead of data for title and message, see ....
```
macdonst: Adding in the `notification` part of the payload will completely screw up the plugin as per [notification vs data payloads](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/PAYLOAD.md#notification-vs-data-payloads). Don't even make a reference to it.

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
...
+}
+```
+
```
macdonst: The information from here down is the key point of this update. I think it needs to be more explicit that you are adding info on how to work with Amazon's SNS service.

docs/PAYLOAD.md
```diff
@@ -72,6 +72,121 @@ Some ways to handle this *double* event are:
...
+    "category": "identifier", // Provide this key with a string value that represents the notification’s type
+    "thread-id": "id", // Provide this key with a string value that represents the app-specific identifier for grouping notifications
+    "sound": "default"  // play default sound, or custom sound, see [iOS Sound](#sound-1) section
+  },
+  "custom_key1": "value1",
+  "custom_key2": "value2"
+}
+```
+
+This is the JSON-encoded format you can send via AWS-SNS's web UI (use "APNS" or "APNS_SANDBOX" depending on which version you use to publish):
```
macdonst: `(use "APNS" to send to an app signed and released to production or "APNS_SANDBOX" or to send to an app signed and released for development)`

docs/PAYLOAD.md
ptandler: Ah, so do you _know_ what the correct location actually is?

ptandler: below the docs say that it should be inside the "data" and named, "content-available": 0 - right?

macdonst:
@ptandler I'm going to merge this then make one quick change as even though some of your examples should be json the comments in it show up in red as they are not valid json. I'll just switch the code blocks to javascript to make the syntax highlighting nicer.

ptandler:
@macdonst thanks for merging and fixing the syntax highlighting

PR #2275, opened 2018-03-21
fix(android): bump play dep

AdriVanHoudt:
## Description
Build broke today, changing this fixes it.
I have no idea why but I thought this might be useful for others.

## Related Issue

https://github.com/phonegap/phonegap-plugin-push/issues/1591 but then v11 to v12 

## Motivation and Context

Fixes our build so I thought it might be useful for others.

## How Has This Been Tested?

Before change --> app instant crashes
After change --> app works fine

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

I think this should only be a bug fix

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2320, opened 2018-04-17
fix(package): update deps

AdriVanHoudt:
## Description
The deps got updated so we need to follow

## Related Issue
https://github.com/phonegap/phonegap-plugin-push/pull/2275

## Motivation and Context
App doesn't boot since there is a mismatch in dep versions

## How Has This Been Tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -60,8 +60,8 @@
       <service android:name="com.adobe.phonegap.push.RegistrationIntentService" android:exported="false"/>
     </config-file>
     <framework src="push.gradle" custom="true" type="gradleReference"/>
-    <framework src="com.android.support:support-v13:23+"/>
-    <framework src="com.google.android.gms:play-services-gcm:12+"/>
+    <framework src="com.android.support:support-v4:24+"/>
+    <framework src="com.google.android.gms:play-services-gcm:15+"/>
     <framework src="me.leolin:ShortcutBadger:1.1.14@aar"/>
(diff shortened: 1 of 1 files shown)


macdonst:
@AdriVanHoudt ug, these dependency issues are a nightmare. The 2.x stream at least is easier to maintain as users can set the values themselves.

jcesarmobile:
@macdonst this PR changed support-v13 to support-v4, I think those are different libraries and this might cause issues.

macdonst:
@jcesarmobile good 👀, I was just setting up to do some testing so I will fix it before releasing.

AdriVanHoudt:
@macdonst yeah they are, and we will once we have some time. In the meantime I hope I can prevent others a bunch of debugging :P
@jcesarmobile I don't know what the diff is between v13 an v4 I just made it work haha

macdonst:
@AdriVanHoudt v4 is when you need to support Android API 4+ and v13 is for 13+.

jcesarmobile:
It no longer works that way, since 26+ they all support API level of 14+, each library implement different things.
But in this case, both 4 and 13 have Fragment classes, so maybe it's not wrong to replace it.
More info about the support libraries: https://developer.android.com/topic/libraries/support-library/packages.html

AdriVanHoudt:
Yeah I took it from here https://github.com/apache/cordova-plugin-camera/blob/d9d80e40c1c44edb0661058dc588c9ea59ff185c/plugin.xml#L84

Issue #2266, opened 2018-03-14
Moving to new user-notifications API of iOS 10+  due to conflict with  katzer/cordova-plugin-local-notification


iliran11:
Hi,
Please see a reference to an issue on the sister repository: https://github.com/katzer/cordova-plugin-local-notifications/issues/1311

on one of the last comments, there is a confirmation: 

> I can confirm that the 0.9.0-beta.1 (katzer/cordova-plugin-local-notifications) running with phonegap-plugin-push@2.1.2 still conflict

Then, the owner, Katzer, explains the reason: 

> The blocker reason is because iOS supports 2 APIs for now. The deprecated API and the new user-notifications API since iOS 10+ however the OS does not allow to use them both together in one app.
> 
> The local-notification plugin uses the new API while the push plugin uses the deprecated API. With the result, that the OS only triggers callbacks for the new API.
> 
> The push plugin has to move to the new API for iOS 10+ otherwise the plugins cannot work side by side.

Is this transition to the new API has happened, or is it still a work to be done?

amit-lulla:
@macdonst : Any ETA on this? Meanwhile, can you suggest if there is a specific version that is stable when used in combination with local-notifications plugin?

iliran11:
it's not about being "stable" version. 
this plugin still Uses the old API of iOS.

your only "option" is to move back to `katzer/cordova-plugin-local-notifications@0.8.4` and lose support of iOS 10.

macdonst:
I have not had the time to look into it. Pull requests are always welcome though.

stellmaurer:
@macdonst This is a huge issue and in my opinion should be top priority. All Ionic apps can't use push notifications and local notifications together. If you could work with @Katzer (develops the local notifications plugin) to fix this issue, that would be awesome.

kadza:
[...]

PR #2363, opened 2018-05-10
Issue #2266:Update to UNUserNotification API of iOS 10+

merrygobyebye:
Update to UNUserNotification API of iOS 10+

## Description
In order to avoid loss of functionality from using deprecated APIs, update the following Apple APIs:

- UILocalNotification => UNNotificationRequest
- UIMutableNotificationAction =>UNNotificationAction
- UIMutableUserNotificationCategory => UINotificationCategory
- UIUserNotificationAction => UNNotificationAction
- UIUserNotificationCategory => UNNotificationCategory
- UIUserNotificationSettings => UNNotificationSettings
- application:handleActionWithIdentifier:forRemoteNotification:completionHandler: => userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:
- application:didReceiveRemoteNotification: => userNotificationCenter:willPresentNotification:withCompletionHandler:

## Related Issue
Issue #2266. It is important to note that this does not solve issue 2266. A further problem that needs to be solved is that [(UNUserNotificationCenter *)currentNotificationCenter](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649510-currentnotificationcenter?language=objc) can only have one delegate. Either the delegate object from cordova-plugin-local-notifications or from this plugin will end up being currentNotificationCenter's delegate, but not both, i.e. only one plugin will get the delegate callbacks.

## Motivation and Context
This change is required because:

Eventually the deprecated Apple APIs will no longer work.

## How Has This Been Tested?

- [x] When sending a message while the app is in the foreground, the 'notification' event is received and the notification is not shown in the system tray.
- [x] When sending a message while the app is in the background, the notification is shown in the system tray.
- [x] When sending a message with content-available: 1 while the app is in the background, the 'notification' event is received and the notification is shown in the system tray. The 'notification' event is received again when the application is opened via tapping the notification.
- [x] When sending a message after killing the app, the notification is shown in the system tray. The 'notification' event is received when the application is opened via tapping the notification.

Environment:
Ionic@3.20.0
Cordova ios@4.5.4
iPhone 8 iOS@11.1.2

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/ios/AppDelegate+notification.h
+++ b/src/ios/AppDelegate+notification.h
@@ -7,13 +7,14 @@
 //
 
 #import "AppDelegate.h"
+@import UserNotifications;
 
-@interface AppDelegate (notification)
+@interface AppDelegate (notification) <UNUserNotificationCenterDelegate>
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken;
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error;
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:( void (^)(UIBackgroundFetchResult))completionHandler;
 - (void)pushPluginOnApplicationDidBecomeActive:(UIApplication *)application;
-- (void)application:(UIApplication *)application handleActionWithIdentifier:(NSString *)identifier forRemoteNotification:(NSDictionary *)userInfo completionHandler:(void(^)())completionHandler;
+- (void)checkUserHasRemoteNotificationsEnabledWithCompletionHandler:(nonnull void (^)(BOOL))completionHandler;
 - (id) getCommandInstance:(NSString*)className;
 
 @property (nonatomic, retain) NSDictionary  *launchNotification;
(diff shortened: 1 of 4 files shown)


knight9999:
I have reviewed this PR and I have noticed that the `completionHandler(UIBackgroundFetchResultNewData)` is always called even when `silent == 1`. 
I think that this method should not be called when `silent == 1`. When `silent == 1`, the completionHandler in the safeHandler should be called. ( `push.finish` method in app fires safeHandler)

merrygobyebye:
@knight9999, from the documentation (https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc), "You must call this handler and should do so as soon as possible." This handler is used to tell the system that the application is done performing its work and may safely be suspended when necessary.

macdonst:
@merrygobyebye yeah, you should do so as soon as possible but once you are done the work. Once the completion handler is called the app will go back to sleep. This is why that completionHandler is called from `push.finish`. It give the JS side a chance to work on the notification.

@knight9999 is going to send a PR on this topic and we can all discuss review it. 

Thanks for sending this PR, it's been weighing on my mind for quite awhile that I hadn't got a chance to do this same work. I really appreciate you taking the time to implement it and I didn't want to come back to you to make changes a month after submitting.

Please look out for @knight9999 PR so we can all make this plugin better.

merrygobyebye:
@macdonst, that makes sense now, thanks for clarifying. 

I'll be sure to review @knight9999's pull request.

knight9999:
Thank you, @merrygobyebye  and @macdonst !
I created a PR. https://github.com/phonegap/phonegap-plugin-push/pull/2417

Issue #2371, opened 2018-05-17
Clearing all notifications remotely (via a stealth notification) Android/iOS


Moghul:
### Expected Behaviour
You send a notification with a "data" payload containing a "clearAllNotifications": "true" key, and the existing notifications disappear.
### Actual Behaviour
N/A
### Reproduce Scenario (including but not limited to)
N/A
#### Steps to Reproduce 
N/A
#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Any platform where code may be run when a notification is received.
#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
N/A
#### Cordova CLI version and cordova platform version

N/A

#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   # e.g. 1.5.3

Developed and tested on 2.1.2

#### Sample Push Data Payload

```
{
    "priority" : "high",
    "content_available" : true,
    "data" : {
      "notId": 0,
      "clearAllNotifications" : "true"
    },
    "to" : "id"
}
```

#### Sample Code that illustrates the problem
See PR when it's made
#### Logs taken while reproducing problem
N/A

What do I want?
I want to be able to remotely trigger the removal of all existing notifications via a hidden notification.

Why do I want it?
Because my notifications are time-sensitive and they become irrelevant after a couple hours.

Why not do it in the "Send custom payload -> pass it to JS -> have JS call plugin to clear all notifications" workflow?
It seems convoluted to me to do it that way and it introduces multiple points of failure. The code to clear notifications must be native, so why not just make it a feature?

I'll have a PR attached to this ASAP so you can take a look at the code.
Note that I don't know objective-c so I just made a thing that works but to which you'll likely cringe. Java is likely easy to improve as well.

Moghul:
@macdonst this ought to work for background ios apps too. Just not force-closed.

PR #2372, opened 2018-05-17
Remotely clear all notifications

Moghul:
Added handling for a `"clearAllNotifications": "true"` key-value pair on iOS and Android which clears the notifications.

## Description

Simply added code which gets the value from the payload, and if it's "true", it executes code which removes notifications from the notification area.

## Related Issue

#2371
## Motivation and Context

It's a straight-forward way to clear time-sensitive notifications an app may have. Simply send the payload, and it's done.

## How Has This Been Tested?

Android 6.0.1 and iOS 11.3.1
Notifications sent via Postman

Sent notifications to both platforms, both in background and force-closed. Then, sent the payload which clears the notififications.

**Android**
Background: Notifications all disappear.
Force-closed: Notifications all disappear.

**iOS**
Background: Notifications all disappear
Force-closed: Nothing happens. Doesn't seem to be able to run the code. Would love some help with this if you want.

**iOS note** : Why does the iOS code first increment the iconApplicationBadgeNumber and then clear it before removing the notifications?
Simply put, it just didn't work otherwise. If you know better, let me know.

Notification used to clear the pre-existing notifications:
```
{
    "priority" : "high",
    "content_available" : true,
    "data" : {
      "notId": 0,
      "clearAllNotifications" : "true"
    },
    "to" : "id"
}
```

In the case that you send a notification with a body and a title, and it also contains this payload (see what I mean below), first, the old notifications will be cleared, and then the new notification will be added. This will result in one remaining notification in the notification drawer.

Android:
```
{
    "priority" : "high",
    "content_available" : true,
    "data" : {
      "title": "some title or another",
      "body": "some body or another",
      "notId": 0,
      "clearAllNotifications" : "true"
    },
    "to" : "id"
}
```

iOS:
```
{
    "priority" : "high",
    "content_available" : true,
    "notification": {
         "title": "some title or another",
         "body": "some body or another",  
    },
    "data" : {
      "notId": 0,
      "clearAllNotifications" : "true"
    },
    "to" : "id"
}
```

N/A
## Types of changes

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/FCMService.java
+++ b/src/android/com/adobe/phonegap/push/FCMService.java
@@ -319,6 +319,7 @@ public class FCMService extends FirebaseMessagingService implements PushConstant
     String title = extras.getString(TITLE);
     String contentAvailable = extras.getString(CONTENT_AVAILABLE);
     String forceStart = extras.getString(FORCE_START);
+    String clearNotifications = extras.getString(CLEAR_ALL_NOTIFICATIONS);
     int badgeCount = extractBadgeCount(extras);
     if (badgeCount >= 0) {
       Log.d(LOG_TAG, "count =[" + badgeCount + "]");
@@ -329,6 +330,12 @@ public class FCMService extends FirebaseMessagingService implements PushConstant
     Log.d(LOG_TAG, "title =[" + title + "]");
     Log.d(LOG_TAG, "contentAvailable =[" + contentAvailable + "]");
     Log.d(LOG_TAG, "forceStart =[" + forceStart + "]");
+    Log.d(LOG_TAG, "clearNotifications =[" + clearNotifications + "]");
+
+    if ("true".equals(clearNotifications)) {
+      NotificationManager notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+      notificationManager.cancelAll();
+    }
 
     if ((message != null && message.length() != 0) || (title != null && title.length() != 0)) {
 
--- a/src/ios/AppDelegate+notification.m
+++ b/src/ios/AppDelegate+notification.m
@@ -107,6 +107,15 @@ static char coldstartKey;
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
     NSLog(@"didReceiveNotification with fetchCompletionHandler");
 
+    id clearAllNotifications = [userInfo objectForKey:@"clearAllNotifications"];
+    if ([clearAllNotifications isKindOfClass:[NSString class]] && [clearAllNotifications isEqualToString:@"true"]) {
+        NSLog(@"PushPlugin clearing notifications");
+
+        [[UIApplication sharedApplication] setApplicationIconBadgeNumber:1];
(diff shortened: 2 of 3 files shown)


fredgalvao:
One point of possible confusion I see is the fact that the handling of the new attribute doesn't define an independent top level behavior, but instead define a new "minor feature", parallel to the usual notification handling. In other terms, you mentioned this in the issue, but we need to make sure this is explicit and clean in the docs, at the very least:
> In the case that you send a notification with a body and a title, and it also contains this payload (see what I mean below), first, the old notifications will be cleared, and then the new notification will be added. This will result in one remaining notification in the notification drawer.

Instead of calling this "a notification payload do clear the drawer", we should call it "a normal notification, that can also explicitly ask to clear the drawer before being handled".

I hope I made myself clear, words are hard.

Moghul:
I understand what you mean. I'll try and whip up something inteligible for the documentation when I have a bit of time.

src/android/com/adobe/phonegap/push/FCMService.java
```diff
@@ -329,6 +330,12 @@ private void showNotificationIfPossible(Context context, Bundle extras) {
     Log.d(LOG_TAG, "title =[" + title + "]");
     Log.d(LOG_TAG, "contentAvailable =[" + contentAvailable + "]");
     Log.d(LOG_TAG, "forceStart =[" + forceStart + "]");
+    Log.d(LOG_TAG, "clearNotifications =[" + clearNotifications + "]");
+
+    if ("true".equals(clearNotifications)) {
```
macdonst: Users may send clearNotifications as a string or as a boolean so we should expect them to mess up and check string equality for "true" and if it is the boolean `true`.

src/ios/AppDelegate+notification.m
```diff
@@ -107,6 +107,15 @@ - (void) application:(UIApplication *)application didReceiveRemoteNotification:(
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
     NSLog(@"didReceiveNotification with fetchCompletionHandler");
 
+    id clearAllNotifications = [userInfo objectForKey:@"clearAllNotifications"];
+    if ([clearAllNotifications isKindOfClass:[NSString class]] && [clearAllNotifications isEqualToString:@"true"]) {
```
macdonst: Same thing as the Android comment. Should check string or boolean values.

```
if (([clearAllNotifications isKindOfClass:[NSString class]] && [clearAllNotifications isEqualToString:@"true"]) || [clearAllNotifications boolValue])
```

src/android/com/adobe/phonegap/push/FCMService.java
Moghul: I was unable to retrieve the value as a boolean even when sent with `"clearAllNotifications": true`. Java claimed it was a string every time, so this is why I set it to a string.

Any ideas why?

macdonst: Not without looking into it in more detail but I'd bet it is some kinda of screwy JSON serialization  issue on Android.

Moghul: I've made some of the changes but I'll leave this one for later.

macdonst:
Superseded by #2521

Issue #2436, opened 2018-06-21
Not getting registration id on second launch (iOS)


jcesarmobile:
### Expected Behaviour
Always get registration id
### Actual Behaviour
Only getting registration id on first launch when accepting push permission
### Reproduce Scenario (including but not limited to)
create an app, add the plugin from github and call `PushNotification.init` and listen for push.on('registration')
#### Steps to Reproduce 
create an app, add the plugin from github and call `PushNotification.init` and listen for push.on('registration')

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
only tested on iOS 10.3.3

#### Cordova CLI version and cordova platform version

    cordova --version                                 8.0.0
    cordova platform version ios               ios 4.5.4

#### Plugin version
master branch (shown as 2.2.3)

#### Sample Push Data Payload
not applicable 

#### Sample Code that illustrates the problem
```
push.on('registration', function (data) {
    console.log(data.registrationId);
    console.log(data.registrationType);
});
```

#### Logs taken while reproducing problem
there is no log about the registration id neither on Xcode console

macdonst:
Merged into master will be part of 3.0.0

PR #2438, opened 2018-06-21
Call registerForRemoteNotifications even if isRegisteredForRemoteNotifications

jcesarmobile:
## Description

registerForRemoteNotifications is only called the first time as it's checking isRegisteredForRemoteNotifications. If registerForRemoteNotifications is not called, then you won't get the token (or failure) again.

## Related Issue
#2436 

## Types of changes

## Checklist:

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -678,9 +678,7 @@
 
 - (void)registerForRemoteNotifications
 {
-    if (![[UIApplication sharedApplication] isRegisteredForRemoteNotifications]) {
-        [[UIApplication sharedApplication] registerForRemoteNotifications];
-    }
+    [[UIApplication sharedApplication] registerForRemoteNotifications];
 }
 
 @end


macdonst:
Thanks @jcesarmobile

merrygobyebye:
@jcesarmobile, @macdonst, perhaps you didn't notice, but `registerForRemoteNotifications` is not only called the first time. See https://github.com/phonegap/phonegap-plugin-push/blob/11d7a6c7db05b1c68769a11fb922c35258a5b720/src/ios/PushPlugin.m#L288 and https://github.com/phonegap/phonegap-plugin-push/blob/11d7a6c7db05b1c68769a11fb922c35258a5b720/src/ios/AppDelegate%2Bnotification.m#L178. It is called:
1. Every time the push plugin is initialized, given the user has notification settings enabled and has not successfully registered for remote notifications.
2. Every time the application becomes active after the push plugin has been initialized, given the user has notification settings enabled and has not successfully registered for remote notifications. 

Were you able to reproduce a scenario in which registering for push notifications never completed successfully? Or is there some desire to receive the registration event more than necessary?

jcesarmobile:
See #2436, I was only getting the token the first app launch, every other launch won't get it, because to receive the token you have to call `[[UIApplication sharedApplication] registerForRemoteNotifications];`.

The token might change, so you have to call `[[UIApplication sharedApplication] registerForRemoteNotifications];` at least on fresh app launch, to make sure the token didn't change.

Not sure if it's necessary to call it on `ApplicationDidBecomeActive`, but checking `[[UIApplication sharedApplication] isRegisteredForRemoteNotifications]` is not good enough as it will always return yes after the fist app launch once the user accept the push prompt, so `[[UIApplication sharedApplication] registerForRemoteNotifications];` won't be called again and the user we won't receive the token ever again.

merrygobyebye:
@jcesarmobile, interesting, I was under the impression that it would be called similarly to the Android method `onTokenRefresh()`, but from the documentation (https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application?language=objc), it's not completely clear.

The only reason I added the call to it on `applicationDidBecomeActive` is to avoid registering for remote notifications until the user grants notification permission via the system dialog or system settings. But it's probably okay to register even if the user hasn't granted notification permission.

jcesarmobile:
This switch is supposed to handle that:
https://github.com/phonegap/phonegap-plugin-push/blob/master/src/ios/PushPlugin.m#L653-L675

If the authorization is `UNAuthorizationStatusNotDetermined` then it will prompt the user, and only register if the permission is granted. It has a block, so the code shouldn't be executed until the user accept or deny the prompt.
If the authorization is `UNAuthorizationStatusAuthorized` will register again.
If the authorization is `UNAuthorizationStatusDenied` it does nothing.

The `applicationDidBecomeActive` part is still useful in case the user accepted when prompted but then disabled the push on setting. Or didn't accept at first but then changed he's mind and enabled in settings.

PR #2509, opened 2018-08-02
Delete FCM instanceID when unregistering - iOS

malwatte:
## Description
FCM Instance ID will be deleted when unregistering. Added some code to check registered with FCM and delete the token if so.

## Related Issue
#2082 

## Motivation and Context
I need to refresh FCM token in certain scenarios. But unregistering and registering again returns the same token. But with this change FCM instance ID will be deleted when unregistered. If you try to send a push notification after unregistering, you will get an "Not registered" error from fcm end-point (which should be the correct behavior I guess). 

## How Has This Been Tested?
Tested on a iPhone 6 (OS version 11.4).
Positive scenario:
Registered/unregistered multiple times. Got a new ID every time as expected. Sent notifications to all tokens. Notifications to the active token worked. Rest failed with "Not registered".
Negative scenario:
error callback was called if no network connection during unregistering.

##Note
Please review the pull request carefully since I am a Objective C noob.

### DIFF
--- a/src/ios/PushPlugin.m
+++ b/src/ios/PushPlugin.m
@@ -119,10 +119,20 @@
             NSLog(@"unsubscribe from topic: %@", topic);
             [pubSub unsubscribeFromTopic:topic];
         }
+    } else if ([self usesFCM]) {
+        [[FIRInstanceID instanceID] deleteIDWithHandler:^void(NSError *_Nullable error){
+            if (error) {
+                // failed to delete the identity for the app
+                [self failWithMessage:command.callbackId withMsg:@"" withError:error];
+            } else {
+                // deleted the identity for the app successfully
+                [self successWithMessage:command.callbackId withMsg:@"unregistered"];
+            }
+        }];
     } else {
         [[UIApplication sharedApplication] unregisterForRemoteNotifications];
         [self successWithMessage:command.callbackId withMsg:@"unregistered"];
-    }
+    }  
 }
 
 - (void)subscribe:(CDVInvokedUrlCommand*)command;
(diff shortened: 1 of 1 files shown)


src/ios/PushPlugin.m
```diff
@@ -119,7 +119,16 @@ - (void)unregister:(CDVInvokedUrlCommand*)command;
             NSLog(@"unsubscribe from topic: %@", topic);
             [pubSub unsubscribeFromTopic:topic];
         }
-    } else {
```
macdonst: Instead of adding an `else if` what about putting that code inside the `else` statement? So…

```
    else {
        if ([self usesFCM]) {
            [[FIRInstanceID instanceID] deleteIDWithHandler:^void(NSError *_Nullable error){
                if (error) {
                    [self failWithMessage:command.callbackId withMsg:@"" withError:error];
                } else {
                    [self successWithMessage:command.callbackId withMsg:@"unregistered"];
                }
            }];
        }
        [[UIApplication sharedApplication] unregisterForRemoteNotifications];
        [self successWithMessage:command.callbackId withMsg:@"unregistered"];
    }
```

This way the APNS token will get released as well.

malwatte: I tested it and there is a small problem. Success callback gets called even when FCM token delete failed. So I did something like this,
```
} else if ([self usesFCM]){
        [[FIRInstanceID instanceID] deleteIDWithHandler:^void(NSError *_Nullable error){
            if (error) {
                [self failWithMessage:command.callbackId withMsg:@"" withError:error];
            } else {
                [[UIApplication sharedApplication] unregisterForRemoteNotifications];
                [self successWithMessage:command.callbackId withMsg:@"unregistered"];
            }
        }];
    }
    else {
        [[UIApplication sharedApplication] unregisterForRemoteNotifications];
        [self successWithMessage:command.callbackId withMsg:@"unregistered"];
    }
```

malwatte: @macdonst ,
Are the changes ok?

Issue #2520, opened 2018-08-09
Suggestion: cancelAll if Android id badge count is zero


chriscant:
### Expected Behaviour

We have a use case which is probably fairly common where notification(s) need to removed when an action has been completed elsewhere, ie a work item has been done on the web so the app notification had best be removed.  A bit like: you've read a Facebook chat message in a browser so it can be cleared from the app.

I expect this to happen even when the app isn't running (even in the background).

My app code makes it work as intended when running in foreground or background ie it can see when count==0 and call clearAllNotifications(). However it doesn't work as expected when my app isn't running at all.

### Actual Behaviour

It works as expected on iOS but doesn't on Android.

#### Steps to Reproduce 

Send a notification with count=1 and a title say.
Sending a further notification with count=0 removes the notification in iOS but not Android.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)

Android 8.1 and 6.

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)

Seen on Nexus 5X and Alcatel Pixi.

#### Cordova CLI version and cordova platform version

Cordova 8.0.0
Cordova android 6.3.0

#### Plugin version

I've reproduced the problem and found workarounds in the plugin 1.11.1 (GCM) and 2.1.3 (Firebase).

#### Sample Push Data Payload

var message = {
    data: {
      title: "All work done",
      message: "date-time"
      count: count.toString(),
      image: "www/img/something.png",
      "content-available": '1',
      notId: Math.floor(new Date().getTime() / 1000).toString()
    },
    apns: {
      payload: {
        aps: {
          badge: count,
        },
      },
    },
    token: token
  };

### Suggested fix

I've tested additions in the function showNotificationIfPossible() in both 1.11.1 GCMIntentService.java and 2.1.3 FCMService.java

```
if (badgeCount == 0) {
  NotificationManager mNotificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
  mNotificationManager.cancelAll();
  return;
}

```
Note that badgeCount will be -1 if a count is not given

It may be that more generally developers might only want the notification cleared if the text is empty as well.

macdonst:
@chriscant seems like a great and simple addition for a PR. Could you send one so we can include it in 3.0.0?

chriscant:
@macdonst will do - probably tomorrow

chriscant:
My suggested code as it stands stops the background notification event handler being called.
[...]

PR #2521, opened 2018-08-10
clearAll notifications on Android if count zero when in background

chriscant:
## Description

Update FCMService notification handler for Android when in background so count==0 initially clears all existing notifications, closely matching iOS functionality.  Note that iOS only clears notifications if the count has gone down eg from 1 to 0. Notifications are not cleared if count not given.

Update payload documentation with note to explain behaviour on Android and iOS.
Please change or move the note if required.

## Related Issue

https://github.com/phonegap/phonegap-plugin-push/issues/2520

## Motivation and Context

If notification(s) are no longer relevant then they should be removed, even if app not running.
This can be signalled by setting badge count to zero.

This was not happening in Android.

Note that after all current notifications have been cleared (a) a new notification will be created if a message or title is given and (b) the notification event handlers are called if the app is running.

## How Has This Been Tested?

Tested by sending notifications with different payloads to Android and iOS devices.

## Types of changes
The change could be viewed as one or all of these types:
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/FCMService.java
+++ b/src/android/com/adobe/phonegap/push/FCMService.java
@@ -329,6 +329,10 @@ public class FCMService extends FirebaseMessagingService implements PushConstant
       Log.d(LOG_TAG, "count =[" + badgeCount + "]");
       PushPlugin.setApplicationIconBadgeNumber(context, badgeCount);
     }
+    if (badgeCount == 0) {
+      NotificationManager mNotificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+      mNotificationManager.cancelAll();
+    }
 
     Log.d(LOG_TAG, "message =[" + message + "]");
     Log.d(LOG_TAG, "title =[" + title + "]");
--- a/docs/PAYLOAD.md
+++ b/docs/PAYLOAD.md
@@ -60,6 +60,8 @@ The following flowchart attempts to give you a picture of what happens when a pu
 - If the app is running in the background the push plugin then checks to see if `content-available` exists in the push data.
 - If `content-available` is set to `1`, then the plugin calls all of your `notification` event handlers.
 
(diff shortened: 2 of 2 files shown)


src/android/com/adobe/phonegap/push/FCMService.java
```diff
@@ -329,6 +329,10 @@ private void showNotificationIfPossible(Context context, Bundle extras) {
       Log.d(LOG_TAG, "count =[" + badgeCount + "]");
       PushPlugin.setApplicationIconBadgeNumber(context, badgeCount);
     }
+    if (badgeCount == 0) {
+      NotificationManager mNotificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+      mNotificationManager.cancelAll();
+    }
```
macdonst: @chriscant okay, I haven't had a chance to test yet but what I see looks good. Let me confirm my thinking.

if count >= 0 then set the badge number, which is the current behaviour
if count == 0 then clear all the notifications which is new behaviour
if count is not provided in the push payload then extractBadgeCount returns -1 and neither of the above two statements are true so we get the existing behaviour.

So the only time the behaviour of the plugin changes is if someone explicitly sets the count/badge to 0 in the push payload? Note, that seems to be the best way forward.

chriscant: That's my understanding, yes.

Note that if a title or message is also supplied then a new notification is created, as per current behaviour.

And the event handler is still called, as per current behaviour.

macdonst: Sweet that's what I was hoping for 🤞 I'll be able to test soon.

macdonst:
@chriscant thanks for the PR. I finally got a chance to test and merge.

PR #2367, opened 2018-05-15
Remove the setting applicationId code

jcesarmobile:
## Description

applicationId is set since cordova-android 5.1.0, so this code is not needed

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2567, opened 2018-09-28
FIX google play services conflicts with the other plugins

SAMIBETTAYEB:
FIX the google play services conflicts with the other plugins like the cordova google maps plugin


push.gradle
```diff
@@ -0,0 +1,34 @@
...
+    def pattern = Pattern.compile(name + "=\"(.*?)\"")
+    def matcher = pattern.matcher(manifestFile.getText())
+    matcher.find()
+    return matcher.group(1)
+}
+
+android {
+    defaultConfig {
+        multiDexEnabled true
+        applicationId = doExtractStringFromManifest("package")
```
fredgalvao: This breaks/conflicts with a custom-set `applicationId` from the user (at the `/platforms/android` level). Do we really need to do this? What's the goal of this `push.gradle` file?

SAMIBETTAYEB: Why does this "break/conflict"?
Yes we need this to improve the build speed.

fredgalvao: Even though the default value for `applicationId` is the name of the package from AndroidManifest, it won't always/necessarily be that. One of my applications, for example, keeps a constant value for the package, while has multiple values for the `applicationId` so that I can maintain custom versions of the same application. I haven't tested it yet on my case, but I'd be willing to say that could conflict with scenarios like mine where the `applicationId` is customized. I could be wrong, but it's definitely a danger-sign for me.

On another topic, how exactly does this gradle file improve build speed?

SAMIBETTAYEB: We're extracting the android package name from the Manifest file, so in all the cases this should always be the same (each time the file should be updated by cordova files generation), In your case you can change only the package name globally and It should be working 👍.
The android build is based on dex, so the multidex enabled should improve the build speed ✌️

fredgalvao:
Fixing conflicts between google play services versions is a welcome and obviously good change, thanks!

However, I can't agree with the overall state of this PR:
1. changing the `applicationId` seems to have no relation at all with the goal of the PR (fix version conflicts) and also no relation at all with the secondary goal of the PR (improve build speed). it does however impose possible conflicts or redundant changes to applications that use a value for `applicationId` different than that of the package from AndroidManifest.
2. Enabling multidex doesn't seem like a trivial and universal change as this PR seems to imply, and doesn't seem to have the positive impact on build speed as mentioned without more complex changes.
According to [the official docs on it](https://developer.android.com/studio/build/multidex), there is the concept of "legacy multidex support", which seems to be universally slower than the normal build: see [this](https://medium.com/mindorks/lets-speed-up-your-gradle-build-6c0e401883e2), [this](https://stackoverflow.com/questions/30177495/gradle-android-studio-build-too-slow-multidex-application), [this](https://hackernoon.com/how-to-speed-up-your-slow-gradle-builds-5d9a9545f91a), and [this](https://android.jlelse.eu/android-build-speed-up-ver-english-c76890aa610). In all of those links, it's given that legacy multidex support (multidex bellow minSdkVersion=21) is slower, and it's given that if you're not on legacy mode (minSdkVersion >= 21), then it can be faster than normal (and the merit is to the incremental build and pre-dex from new android build tools versions, not multidex itself). 

There's a very good explanation/summary of those issues and the not-so-trivial workaround for it [here](https://learnpainless.com/android/how-to-speed-up-gradle-build-time-also-if-multidex-enabled).

And even not considering the actual slowdown on build speed caused by multidex on legacy mode, the docs mention that additional changes are needed to enable it on minSdkVersion<21 (there's a required dependency to make it work), and the applications that enable multidex on legacy support without that dependency would break.

Considering this plugin's latest version (and master branch) support `cordova-android@7.1.0` ([source](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/INSTALLATION.md#installation-requirements)), and that said version supports sdk version from 19 to 27 ([source](https://cordova.apache.org/docs/en/latest/guide/platforms/android/#requirements-and-support)) (meaning that most applications on the latest version of cordova-android will still be supporting sdk versions that trigger the legacy multidex support mode), I'd conclude that the chances of this change breaking a lot of applications are big.


All of the above was taken from theory/docs or from other people's experience only, and I have no real-world experience with multidex builds (I never had to enable it), so **please correct me if I'm wrong**.

Of course this is only one perspective, @macdonst or @jcesarmobile 's takes on this are needed.

push.gradle
jcesarmobile: The `applicationId` is set since cordova-android 5.1.0, so you don't really need to set the value based on the AndroidManifest.xml value, it should always match now. I recently removed this code in https://github.com/phonegap/phonegap-plugin-push/pull/2367
And the multidex is still applied, but by using a dependency plugin https://github.com/phonegap/phonegap-plugin-push/blob/master/plugin.xml#L47

So this push.gradle file is not needed

plugin.xml
```diff
@@ -39,12 +39,13 @@
       </service>
     </config-file>
     <preference name="ANDROID_SUPPORT_V13_VERSION" default="27.+"/>
-    <preference name="FCM_VERSION" default="17.0.+"/>
+    <preference name="FCM_VERSION" default="17.3.2"/>
```
jcesarmobile: I think google added breaking changes on 17.1.x, so updating to 17.3.2 will probably break the plugin unless we do some code modifications.

jcesarmobile:
Added a few comments, but I'm going to close the PR because updating the default version doesn't really fix conflicts, the version is configurable, so if you want to use 17.3.x you can configure that version on install time with a variable, no need to update the default. That's the whole point of being configurable.
But be careful, because as I commented, 17.1.x+ will probably break plugin code.

SAMIBETTAYEB:
@jcesarmobile, the problem is not about the configurable variables, I know that we can set the FCM one, the problem here, is that the firebase messages, and firebase core or maybe analytics for example have different versions for the other plugins, they don't have the same version on the same release, for example, for example the 17.3.2 for the messages and the 16.0.4 for the core, if you set the variable to 17.3.2, the problem is that the core has no version 16.0.4, if we set the version to 16.0.4, the same thing for the messages that has no version for the 16.0.4.
In this case, we need to define a variable for each service, or to define them manually.

jcesarmobile:
Yeah, but I don’t see how your PR fixes that, it just changes default version

Issue #2297, opened 2018-04-03
Typescript types are out of date


oliverjanik:
Types here: 

https://github.com/phonegap/phonegap-plugin-push/blob/master/types/index.d.ts

require SenderID which does not exist according to the docs here:

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md

fredgalvao:
@oliverjanik Thanks for the detail.
This is a duplicate of https://github.com/phonegap/phonegap-plugin-push/issues/1536, will close this and focus on that one.

PR #2596, opened 2018-10-22
Update TYPESCRIPT.md

simonhaenisch:
## Description

The type declarations for this plugin come with the package now, [`@types/phonegap-plugin-push`](https://www.npmjs.com/@types/phonegap-plugin-push) has been deprecated.

The example also didn't show how to actually import the types.

## Related Issue

Nope.

## Motivation and Context

I needed to use this plugin in a Typescript project and saw that this doc was outdated.

## Types of changes

## Checklist:

### DIFF
--- a/docs/TYPESCRIPT.md
+++ b/docs/TYPESCRIPT.md
@@ -1,7 +1,6 @@
-# Typescript definition file
+# Typescript definitions
 
-For those of you who use typescript, we're glad to say that we have the complete definition file available at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).
-Search for `phonegap-plugin-push` there, or simply grab it directly [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/phonegap-plugin-push/index.d.ts).
+For those of you who use Typescript, we're glad to say that we provide the complete definition file along with our package.
 
 ## Example usage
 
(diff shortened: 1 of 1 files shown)


docs/TYPESCRIPT.md
```diff
@@ -64,8 +65,3 @@ push.on('notification', (data: my.custom.NotificationEventResponse) => {
 	console.log(data.additionalData.bacon);
 });
 ```
-
-## Outdated definitions
-
-Is our definition file at DefinitelyTyped outdated? Is there any improvements that could be done?
-We welcome any contribution, and they should be done through issues created [there](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/new).
```
fredgalvao: I'd say we're better off just switching `and they should be done through issues created *here*` instead of removing this part, even though I agree that it's kinda implied for the whole project/repo. What do you say?

simonhaenisch: Was thinking the same first, but now that the declarations are part of the source... so I don't think the definitions should ever be considered outdated again. If there is an issue with the definitions, it's just a bug now, I guess.

fredgalvao: I can agree with that. I'll make room on schedule to bring the definitions up to par with the actual API/source ASAP so that we're good from then on.

fredgalvao:
Thanks for your time and for the contribution!

fredgalvao:
https://github.com/phonegap/phonegap-plugin-push/issues/2297 and https://github.com/phonegap/phonegap-plugin-push/issues/1536 have been the target issue for a while, and I'll link this here for consistency.

PR #2607, opened 2018-11-01
Added Channel Group

OliverOng1995:
## Description
In Android O,Push Channel also have group to categorize notification channel such as Work and Personal Notifications  

## Motivation and Context
It can help user to assign notification channel to a group,easily categorizing it

## How Has This Been Tested?
Using Outsystem,I use the Firebase Push Plugin and modified the source code and the wrapper on the plugin on Outsystem itself with the channels related function in the latest version of this plugin.
Also tested on Mi Max on lineageOS 15.1 and Mi A1 on Android 8.0 for creation,listing and deletion of channel.Setting and listing channels with the group

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My change requires a change to the documentation.
- [ x] I have updated the documentation accordingly.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -5,6 +5,9 @@
 - [.createChannel() - Android only](#pushnotificationcreatechannel)
 - [.deleteChannel() - Android only](#pushnotificationdeletechannel)
 - [.listChannels() - Android only](#pushnotificationlistchannels)
+- [.createChannelGroup() - Android only](#pushnotificationcreatechannelgroup)
+- [.deleteChannelGroup() - Android only](#pushnotificationdeletechannelgroup)
+- [.listChannelGroups() - Android only](#pushnotificationlistchannelgroups)
 - [push.on()](#pushonevent-callback)
   - [push.on('registration')](#pushonregistration-callback)
   - [push.on('notification')](#pushonnotification-callback)
@@ -224,6 +227,7 @@ A default channel with the id "PushPluginChannel" is created automatically. To m
 | `sound`                          | `String`  | The name of the sound file to be played upon receipt of the notification in this channel. Cannot be changed after channel is created.                                                                                               |
 | `vibration`                      | `Boolean` or `Array` | Boolean sets whether notification posted to this channel should vibrate. Array sets custom vibration pattern. Example - vibration: `[2000, 1000, 500, 500]`. Cannot be changed after channel is created.                 |
 | `visibility`                     | `Int`     | Sets whether notifications posted to this channel appear on the lockscreen or not, and if so, whether they appear in a redacted form. 0 = Private, 1 = Public, -1 = Secret.                                                         |
+| `group_id`                     | `String`     | Sets channel group id. Assign it to a Channel Group Created. Cannot be changed once set.                                                         |
 
 ## PushNotification.deleteChannel(successHandler, failureHandler, channelId)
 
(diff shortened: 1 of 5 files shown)


fredgalvao:
Thanks for contributing!

Can you update the type definitions to reflect on those new APIs and attributes?

OliverOng1995:
What is it,is it the API.md

fredgalvao:
It's the https://github.com/phonegap/phonegap-plugin-push/blob/master/types/index.d.ts file. That typescript "interface descriptor" needs to reflect the new APIs you're adding, so that the project compiles for the users that consume our API through a typescript project.

OliverOng1995:
There you have it

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
 		from?: string
 		notId?: string
 	}
-
+  /**
+   * Set Channel Variable,once created cannot be changed programmtically for importance and vibration
+   */
+  interface channel{
```
fredgalvao: This should be upper case, as it's a proper type name.
```suggestion
  interface Channel {
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    vibration: boolean
+    /**
+     * Set notitfication visibility 
+     */
+    visibility: integer
+    /**
+     * Set Channel Group
+     */
+    groupId:string
```
fredgalvao: Are all of the attributes on the `Channel` interface mandatory? Or can the user ommit some of them? If any of them happens to be optional, you can mark it like so:
```suggestion
    groupId?: string
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    name: string 
+    /**
+     * camillebeaumont name and description fix,channel description
+     */
+    description: string
+    /**
+     * channel importance level
+     */
+    importance: integer
```
fredgalvao: The correct typescript type for numbers is `number`, we don't have specialized numeric types in typescript/javascript:
```suggestion
    importance: number
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+    /**
+     * enable vibration for the notification
+     */
+    vibration: boolean
+    /**
+     * Set notitfication visibility 
+     */
+    visibility: integer
```
fredgalvao: Same `integer` vs `number` comment from above.

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+    /**
+     * Set Channel Group
+     */
+    groupId:string
+  }
+  
+   interface channelGroup{
```
fredgalvao: Same naming suggestion from the `Channel` interface applies here:
```suggestion
   interface ChannelGroup {
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+    /**
+     * Set Channel Group ID
+     */
+    id: string
+   /**
+     * Set Channel Group Name
+     */
+    name: string
```
fredgalvao: Same "can these be optional" suggestion applies to these attributes.

types/index.d.ts
fredgalvao: I just realized we don't have these documented [here](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#channel-properties) with regards to:
- are them optional?
- do they have default values?
So I'll take my previous comment back, it's not your responsibility to fix this.

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+  /**
+   * Set Channel Variable,once created cannot be changed programmtically for importance and vibration
+   */
+  interface channel{
+    /**
+     * channel ID
+     */
+    id: string
+    /**
+     * camillebeaumont name and description fix
```
fredgalvao: Are these notes to fix later? We don't have the `name` attribute yet, so we don't have an official doc to copy from, so it would make sense.

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
 		from?: string
 		notId?: string
 	}
-
+  /**
+   * Set Channel Variable,once created cannot be changed programmtically for importance and vibration
+   */
+  interface channel{
+    /**
+     * channel ID
```
fredgalvao: We can copy the docs on this from the official API docs.
In this case, it would be:
```suggestion
     * The id of the channel. Must be unique per package. The value may be truncated if it is too long.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    id: string
+    /**
+     * camillebeaumont name and description fix
+     */
+    name: string 
+    /**
+     * camillebeaumont name and description fix,channel description
```
fredgalvao: In this case we also have official docs to copy from:
```suggestion
     * The user visible name of the channel. The recommended maximum length is 40 characters; the value may be truncated if it is too long.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    description: string
+    /**
+     * channel importance level
```
fredgalvao: Same from above:
```suggestion
     * The importance of the channel. This controls how interruptive notifications posted to this channel are. The importance property goes from 1 = Lowest, 2 = Low, 3 = Normal, 4 = High and 5 = Highest.
```

types/index.d.ts
OliverOng1995: Sorry about that! Not familiar with typescript

types/index.d.ts
OliverOng1995: groupId is optional

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    importance: integer
+    /**
```
fredgalvao: We're missing the `sound` attribute on the `Channel` object interface. We might as well use the opportunity to make the whole interface documented.
```suggestion
    /**
     * The name of the sound file to be played upon receipt of the notification in this channel. Cannot be changed after channel is created.
     */
    sound: string
    /**
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     * enable vibration for the notification
```
fredgalvao: Official docs:
```suggestion
     * Boolean sets whether notification posted to this channel should vibrate. Array sets custom vibration pattern. Example - vibration: [2000, 1000, 500, 500]. Cannot be changed after channel is created.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    vibration: boolean
+    /**
+     * Set notitfication visibility
```
fredgalvao: Official docs:
```suggestion
     * Sets whether notifications posted to this channel appear on the lockscreen or not, and if so, whether they appear in a redacted form. 0 = Private, 1 = Public, -1 = Secret.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+    /**
+     * enable vibration for the notification
+     */
+    vibration: boolean
+    /**
+     * Set notitfication visibility 
+     */
+    visibility: integer
+    /**
+     * Set Channel Group
```
fredgalvao: Let's use the same docs from the API.md table:
```suggestion
     * Sets channel group id. Assign it to a Channel Group Created. Cannot be changed once set.
```

OliverOng1995:
Sorry I am not familiar with typescript,sorry to inconvenience you with my new features that I want to add on

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    groupId:string
+  }
+  
+   interface channelGroup{
+    /**
+     * Set Channel Group ID
```
fredgalvao: Same docs from the API.md:
```suggestion
     * The id of the channel group.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+     */
+    id: string
+   /**
+     * Set Channel Group Name
```
fredgalvao: Same docs from the API.md:
```suggestion
     * The user visible name of the channel group.
```

types/index.d.ts
```diff
@@ -289,10 +289,61 @@ declare namespace PhonegapPluginPush {
...
+  }
+  
 	interface PushNotificationStatic {
 		init(options: InitOptions): PushNotification
 		new (options: InitOptions): PushNotification
+    hasPermission(successHandler: () => any): void 
+    createChannel(successHandler: () => any, errorHandler: () => any,channel): void
+    listChannels(successHandler: () => any): void
+    deleteChannel(successHandler: () => any, errorHandler: () => any, channelId: string ): void
+    createChannelGroup(successHandler: () => any, errorHandler: () => any,channelGroup): void
```
fredgalvao: The `channelGroup` parameter should be explicitly typed as the interface that represents it.
```suggestion
    createChannelGroup(successHandler: () => any, errorHandler: () => any, channelGroup: ChannelGroup): void
```

types/index.d.ts
fredgalvao: So let's have that suggestion back :smile:

types/index.d.ts
OliverOng1995: Yup based on a previous pull request

fredgalvao:
> Sorry I am not familiar with typescript variable,sorry to inconvenience you with my new features that I want to add on.I thought it seemed like javascript

Not at all an inconvenience! Please pat yourself on the back, you're doing all of us a big favor.

You even walked the extra mile to document/type missing parts that weren't even related to your additions to the API, which is awesome.
The simple fact that you're taking the time to work through suggestions on this PR is another amazing things. So yes, thank you very much for your contribution ;)

I won't bother you with some of the other missing parts I noticed, as they're definitely not related to your PR, and I'll work on them later myself.

types/index.d.ts
OliverOng1995: It is not optional

Issue #2617, opened 2018-11-07
Trying to replace notifications on iOS (using `notId`)


wolfenrain:
### Expected Behaviour
That when using the `notId` the notification will be replaced on iOS.

### Actual Behaviour
I just get a new notifiction instead of the old one replaced.

### Reproduce Scenario (including but not limited to)
Send this Android payload using FCM to an Android device:
```json
{
  "delayWhileIdle": false,
  "collapseKey": "testKey",
  "priority": "high",
  "dryRun": false,
  "data": {
    "notId": 851980,
    "color": "#6ECFF6",
    "ledColor": [0, 110, 207, 246],
    "fullscreen": "0",
    "content-available": "1",
    "no-cache": "0",
    "title": "Chat message",
    "body": "Hello",
    },
    "android_channel_id": "channel-chat",
    "priority": 2,
    "soundname": "default"
  }
}
```
This will go exactly as expected, the notification will be received, if you send it again but with a different body, you will get a replaced notification. As expected

Send this iOS payload using FCM:
```json
{
  "delayWhileIdle": false,
  "collapseKey": "testKey",
  "priority": "high",
  "dryRun": false,
  "data": {
    "notId": 851980,
    "color": "#6ECFF6",
    "ledColor": [0, 110, 207, 246],
    "fullscreen": "0",
    "content-available": "1",
    "no-cache": "0"
  },
  "notification": {
    "title": "Chat message",
    "body": "hello",
    "badge": 2,
    "content-available": "1", // I have tested it with and without this(originally it was without)
    "sound": "default.caf"
  }
}
```
You will receive a notification, but when you change the body and send the notification again, you get *another* one instead of it being replaced.

#### Steps to Reproduce 
See Reproduce Scenario.

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Android 8 and iOS 12.1

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
Motorola

#### Cordova CLI version and cordova platform version
cordova: `8.0.1-nightly.2018.8.31.dfd2fe79 (cordova-lib@8.0.1-nightly.2018.8.31.d93cfa77)`
cordova android: `7.1.1`
cordova ios: `4.5.5`

#### Plugin version
2.2.3, but with current master code.

#### Sample Push Data Payload
See Reproduce Scenario.

#### Sample Code that illustrates the problem
test script to easily test(duh)
```
#!/usr/bin/env node

[...]

Issue #2585, opened 2018-10-11
Clear notification based on other notification


wolfenrain:
Hi,
I am building an app which uses the phonegap-plugin-push for notifications, but I want to have to ability to clear notification on a device through my server. As far as I am aware this is currently not possible. But I had the idea to make use of the local `clearNotification` method, if the received notification contains a specific payload, lets says `clearNotificiationId`, and we have that notification id, then clear it. 

This allows developers to clear notification on devices when the app is not running. I am willing to make PR for this, considering I need it myself. But I was wondering what your thoughts are about this before providing it. Or if you know of a way to do this already.

If the given solution is not wanted, or should be implemented differently, I am all ears. If its not wanted I will just make a fork and implemented it myself. But I hope it is possible to keep it in the upstream.

jcesarmobile:
Can't you send the notification id and use it to clear it with `clearnotification` method?

https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/API.md#pushclearnotificationid-successhandler-errorhandler---ios--android-only

wolfenrain:
Well yes ofcourse, but that would only work if the app is not killed, the idea is to make it possible from the background swell.

macdonst:
@wolfenrain sure, I'd love to see a PR on this.

andreszs:
Not even the Gmail app features this.... notifications for read emails persist in the phone notifications bar. The current approach is to simply update the old notification with a newer, updated one, and it's not that bad anyway.

mark-veenstra:
@andreszs Well that is how GMail works. But what if you build a chat application or a phone call application? I can imagine the use-case that people would like to send a notification to a user, notify-ing them that there are unread chats. But the user has multiple devices hooked to the chat application and reads the messages on device 1. On device 2 we would like to completely remove that notification.

Same goes for a phone application. If a user has 2 devices hooked to the application and both are ringing because of a push messages. On 1 device you answer. On the other device you would like to remove the push notification as the call has been answered.

I guess there are use-cases you can imagine that is useful for having this solution.

wolfenrain:
@macdonst I will start working on the PR now. :-)

PR #2619, opened 2018-11-08
Issue 2585 - Alllowing to clear notification when the app is not running

wolfenrain:
***NOTE***: iOS hasn't been tested yet because of #2617. I am delivering the PR nonetheless for discussions and possible changes if needed. When I am able to do the usecase described in #2617 I will be able to test iOS and fix any mistakes if needed.

## Description
I added an extra optional key to the payload `clearNotification`, this key can be used to clear notificitations that have said `notId`.
For example:
```javascript
// Send first notificitation
firstPayload = {
        'title': 'Clear notification',
        'body': 'Test body',
        'notId': '1337
};
// Send second notification to clear it
secondPayload = {
        'title': 'Clear notification',
        'body': 'Test body',
        'clearNotification': '1337'
};
```
The changes are quite simple, for Android I added an extra check if the `clearNotification` key is available, if so then we simple remove the notification with that ID. I did change the `clearNotification()` method. I move the functionality to a static method. That way anyone that has acces can call the method while the cordova bridge is still intact.

For iOS basically the same thing except that I moved the `clearNotification()` functionality to `clearRealNotification()` so it can be used by anyone that has acccsm, and still keeping the cordova bridge intact.

## Related Issue
#2585

## Motivation and Context
This allows developers to remove notification from the notification tray without their user having the app open/running.

And here are some [use-cases](https://github.com/phonegap/phonegap-plugin-push/issues/2585#issuecomment-434958216) for this functionality describe by an other user.

## How Has This Been Tested?
I tested this by sending above payload through FCM towards my devices. Android functionality hasn't changed much, I tested the push and the cordova bridge and both are still working as expected.

iOS testing *HAS NOT BEEN DONE YET* because of #2617. Because the `clearNotification` is in the same method as the `notId`, I haven't been able to test it yet.

Documentation is also not done because I wasn't sure where to put the new information, if anyone can point me in that direction, much appreciated.

## Types of changes

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/FCMService.java
+++ b/src/android/com/adobe/phonegap/push/FCMService.java
@@ -103,8 +103,13 @@ public class FCMService extends FirebaseMessagingService implements PushConstant
         PushPlugin.setApplicationIconBadgeNumber(getApplicationContext(), 0);
       }
 
+      int clearNotificationId = getClearNotificationId(extras);
+
+      // if the notification has a `clearNotification`, then clear it.
+      if(clearNotificationId != -1) {
+        PushPlugin.clearNotification(getApplicationContext(), clearNotificationId);
       // if we are in the foreground and forceShow is `false` only send data
-      if (!forceShow && PushPlugin.isInForeground()) {
+      } else if (!forceShow && PushPlugin.isInForeground()) {
         Log.d(LOG_TAG, "foreground");
         extras.putBoolean(FOREGROUND, true);
         extras.putBoolean(COLDSTART, false);
@@ -300,6 +305,20 @@ public class FCMService extends FirebaseMessagingService implements PushConstant
     } // while
 
     return newExtras;
+  }
+
+  private int getClearNotificationId(Bundle extras) {
+    int clearNotificationId = -1;
+    String msgcnid = extras.getString(CLEAR_NOTIFICATION);
+
+    try {
+      if (msgcnid != null) {
+        clearNotificationId = Integer.parseInt(msgcnid);
+      }
+    } catch (NumberFormatException e) {
+      Log.e(LOG_TAG, e.getLocalizedMessage(), e);
+    }
+    return clearNotificationId;
   }
 
   private int extractBadgeCount(Bundle extras) {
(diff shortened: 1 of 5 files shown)


src/android/com/adobe/phonegap/push/FCMService.java
```diff
@@ -103,8 +103,13 @@ public void onMessageReceived(RemoteMessage message) {
...
       }
 
+      int clearNotificationId = getClearNotificationId(extras);
+
+      // if the notification has a `clearNotification`, then clear it.
+      if(clearNotificationId != -1) {
+        PushPlugin.clearNotification(getApplicationContext(), clearNotificationId);
       // if we are in the foreground and forceShow is `false` only send data
-      if (!forceShow && PushPlugin.isInForeground()) {
+      } else if (!forceShow && PushPlugin.isInForeground()) {
```
fredgalvao: The way this is right now, and considering the example you used on the OP, I can still use the `clearNotification` attribute on the payload to clear said id if the application is in foreground, *however*  the scenario on this line won't trigger and it'll fall through to the last `else {`, which will consider foreground:false even though we're in foreground, therefore creating a very inconsistent payload for the js to handle.

So, we have a few options:
- we can consider that the `clearNotification` attribute represents a separate and independent feature on the payload, and a single payload can be {an actual notification with all the other payload attributes available} and {a request to clear a specific notification} *at the same time*. This would need the first `if` statement here to be isolated as to not race against the existing ones. Check the `clearBadge` feature, as that's the closest we have to this option.
- we can consider that the `clearNotification` attribute represents the whole feature, and whenever it is present it will skip all other cases (`if/else` statements). This would need the first `if` statement to actually return from the method, or make the last `else` statement do nothing if it did clear a notification before. This would also make the example used on the OP invalid, as the `title` and `body` attributes on the payload mean nothing. This would also make it useful to move this feature to as early as possible on the method to avoid useless work (if we're gonna clear a notification and return, why waste resources preping the extras? This would also make it needed to solve the question "who has the highest priority on the feature order: `clearBadge` or `clearNotification`?".

fredgalvao:
Thanks for this!

I left a beefy comment on the code review, btw.
Also, we're gonna need the docs to be updated for this after that comment is resolved.

src/android/com/adobe/phonegap/push/FCMService.java
wolfenrain: I would consider going for the first option, I would consider that the most likely use-case for people. And that way you don't force people. Will change and test it :-)

Issue #2630, opened 2018-11-22
Force-start: 1 app start in background but received next notification in foreground state 


McSam94:
### Expected Behaviour
After force-start, next notification received in background state

### Actual Behaviour
After force-start, next notification received in foreground state

#### Steps to Reproduce 
1) Terminate App from task manager
2) Send Push Notification 
3) App force start in background
4) Send Push Notification
5) App received notification in foreground state

#### Platform and Version (eg. Android 5.0 or iOS 9.2.1)
Android 7.0

#### (Android) What device vendor (e.g. Samsung, HTC, Sony...)
Honor 8

#### Cordova CLI version and cordova platform version
    cordova --version                                     8.0.0
    cordova platform version android                     6.3.0

#### Plugin version

    cordova plugin version | grep phonegap-plugin-push   2.0.0

#### Sample Push Data Payload
           ```
 var notificationData = new notification();
            notificationData.body = messageBody;
            notificationData.title = messageTitle;
            notificationData.sound = "Enabled";
            notificationData.style = "inbox";
            notificationData.summaryText = "%n% Notifications";
            notificationData.type = messageType;
            notificationData.notId = g; //random number
            notificationData.force_start = 1;

               var dataToSend = new
                {
                    registration_ids = receivers,
                    data = notificationData
                };
```

#### Logs taken while reproducing problem
logcat | grep Push
```
11-22 10:01:26.547  8865  8865 I chromium: [INFO:CONSOLE(1)] "10:01:26,545 PushService.ts Push init entry", source: file:///android_asset/www/build/vendor.js (1)
11-22 10:01:26.623  8865  8865 I chromium: [INFO:CONSOLE(1)] "10:01:26,622 PushService.ts Push notifications granted", source: file:///android_asset/www/build/vendor.js (1)
11-22 10:01:26.624  8865  8865 I chromium: [INFO:CONSOLE(1)] "10:01:26,624 PushService.ts initPush entry", source: file:///android_asset/www/build/vendor.js (1)
11-22 10:01:26.624  8865  8865 I chromium: [INFO:CONSOLE(1)] "10:01:26,624 PushService.ts initPush exit", source: file:///android_asset/www/build/vendor.js (1)
11-22 10:01:26.689  8865  8865 I chromium: [INFO:CONSOLE(1)] "10:01:26,688 PushService.ts Subscription to '/topics/all' : OK 3000", source: file:///android_asset/www/build/vendor.js (1)
[...]

PR #2646, opened 2018-12-09
:bug: Issue #2630: Fix initialisation of gForeground when started in …

warthog618:
## Description

Fix initialisation of plugin to handle the case where the app is launched in the background.

## Related Issue

Issue #2630 

## Motivation and Context

When currently launched in the background, the plugin incorrectly thinks it is in the foreground and absorbs notifications instead of sending them to the shade.  This behaviour continues indefinitely until the app is brought into the foreground by the user at which point the app and plugin state get back into sync.

The fix keeps the plugin gForeground state flag in sync with the app at all times.

## How Has This Been Tested?

Tested as described in my comments in Issue #2630:

Using the following msg:
```
fsmsg =
  data:
    'force-start': 1
    title: 'forced'
    message: 'any message'
```
1. With the app not running, send fsmsg notification.
2. Phone receives the notification which is shown in the shade, and the app is launched in the background.
3. Do NOT bring the app to the foreground.
4. Send fsmsg again.  The phone should indicate that the notification has been received (without the fix it doesn't).

## Types of changes

## Checklist:

### DIFF
--- a/src/android/com/adobe/phonegap/push/PushPlugin.java
+++ b/src/android/com/adobe/phonegap/push/PushPlugin.java
@@ -499,7 +499,8 @@ public class PushPlugin extends CordovaPlugin implements PushConstants {
   @Override
   public void initialize(CordovaInterface cordova, CordovaWebView webView) {
     super.initialize(cordova, webView);
-    gForeground = true;
+    boolean startInBackground = cordova.getActivity().getIntent().getExtras().getBoolean(START_IN_BACKGROUND,false);
+    gForeground = !startInBackground;
   }
 
   @Override


warthog618:
> I get the following exception when trying your branch @warthog618 :
> 
> ```
> W/System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.os.Bundle.getBoolean(java.lang.String, boolean)' on a null object reference
>         at com.adobe.phonegap.push.PushPlugin.initialize(PushPlugin.java:502)
>         at org.apache.cordova.CordovaPlugin.privateInitialize(CordovaPlugin.java:57)
> W/System.err:     at org.apache.cordova.PluginManager.getPlugin(PluginManager.java:171)
>         at org.apache.cordova.PluginManager.exec(PluginManager.java:122)
>         at org.apache.cordova.CordovaBridge.jsExec(CordovaBridge.java:59)
>         at org.apache.cordova.engine.SystemExposedJsApi.exec(SystemExposedJsApi.java:41)
>         at android.os.MessageQueue.nativePollOnce(Native Method)
> ```

Well I don't, which makes you more able to debug your problem than me.
Perhaps you could provide more info or debug it yourself???

As my change is essentially a one liner:

`boolean startInBackground = cordova.getActivity().getIntent().getExtras().getBoolean(START_IN_BACKGROUND,false);
`

it appears the getExtras() is returning a null for you, but not for me.

Elsewhere in the plugin code there are checks for getExtras returning null, through not in the section that my line is derived from (PushHandlerActivity.onCreate).  So it may be necessary to check if extras exists.
I'm not sufficiently familiar with how cordova plugins work to determine in what circumstances there are no extras, nor what value startInBackground should default to in that case, so feel free to work out why it isn't working for you, and what the implications are, and suggest a patch - as I did for my problem.

Have a good one.

BorntraegerMarc:
yeah I did debug and it seems that `getExtras` is nullable. So it can be null sometimes.

Best practice seems to be to check it for `null`. Source: https://stackoverflow.com/questions/32031178/nullpointerexception-on-android-os-bundle

BorntraegerMarc:
@warthog618 but I tested your code and for me my app still goes into foreground when I receive 2 push notifications.

warthog618:
@BorntraegerMarc That SO reference is a very specific case that doesn't seem to be relevant to this problem - from the answer:

"This is happen because when you start your second activity NewTransaction directly you don't put extras in the intent, so when you call getIntent().getExtras(); it returns a null object and this is why getIntent().getExtras().getBoolean("update"); throw the NPE."

I don't see the plugin calling NewTransaction anywhere, so that doesn't seem to apply.
The open question is "in what circumstances might the activity intent contain no extras?"
I'm not a plugin developer, so I don't know the answer to that.

Can I suggest that you debug the problem that you are seeing, rather than googling it, and then suggest a patch?  And as I already alluded to, there are other places in the plugin code that don't check for getExtras returning null, so they may also require additional checks unless you can guarantee that extras will exist.

Wrt your latest comment, so you tested the code, got a NPE and then expect it to still work???
And you expect me to workout why it is broken for you?  That is your itch to scratch, not mine.

Happy debugging, and I look forward to your patch.

BorntraegerMarc:
@warthog618 well I didn't want to offend you 😄 I am debugging since some hours 😄 

Since you wrote `Do NOT bring the app to the foreground.` I thought you fixed this bug you described. But since it didn't work for me I thought I write this to you in case you didn't test your changes on a physical device.

Testing on android 9, btw.

warthog618:
@BorntraegerMarc I only test on physical devices.  I did test the fix as I described on physical devices with both Android 7.1.2 and 8.1.0 - the two versions that I had available at the time.
And I have just tested it on Android 9 (specifically LineageOS 16.0) and it is still working for me.
So still no idea what is different for you.

My current build environment is:

```
Ionic:

   ionic (Ionic CLI) : 4.6.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework   : unknown
   @ionic/v1-toolkit : 1.0.15

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : android 7.1.4
   Cordova Plugins       : no whitelisted plugins (17 plugins total)

System:

   Android SDK Tools : 26.1.1 (/Users/kent/Library/Android/sdk)
   NodeJS            : v10.14.2 (/usr/local/bin/node)
   npm               : 6.4.1
   OS                : macOS High Sierra
   Xcode             : Xcode 10.1 Build version 10B61 

```

BorntraegerMarc:
OK, so what I can confirm is that for me the app starts in foreground when:
- push notifications are sent in a short time (between 1 and approx. 8 seconds) AND
- you kill the app at least once when started in background
- The plugin seems to behave correctly if the notifications are at least 9 or 10 seconds apart

I tested this with an empty `OnNotification` handler in our cordova app.

BUT the behaviour is even worse without the fix of this PR. Because then the app opens always (doesn't matter how far apart the push notifications are).

Could somebody merge this PR and release a new version of the plugin? This fix is desperately needed if one wants to do some background work when notifications arrive.

Issue #2703, opened 2019-02-11
IOS: Missing Push Notification Entitlement: Phonegap build error


regevbr:
#### Error Message
Missing Push Notification Entitlement - Your app includes an API for Apple's Push Notification service, but the aps-environment entitlement is missing from the app's signature. To resolve this, make sure your App ID is enabled for push notification in the Provisioning Portal. Then, sign your app with a distribution provisioning profile that includes the aps-environment entitlement. This will create the correct signature, and you can resubmit your app. See "Provisioning and Development" in the Local and Push Notification Programming Guide for more information. If your app does not use the Apple Push Notification service, no action is required. You may remove the API from future submissions to stop this warning. If you use a third-party framework, you may need to contact the developer for information on removing the API.

#### Suggestion
hi, looks like phonegap build has changed or updated something. working due to Missing Push Notification Entitlement error during submission to itune. Push on android is working fine.

regevbr:
```
        <config-file target="*.entitlements" parent="aps-environment">
          <string>production</string>
        </config-file>
```
Should fix it

PR #2704, opened 2019-02-11
#2703 IOS: Missing Push Notification Entitlement

regevbr:
add missing entitlements

## Description
add missing entitlements

## Related Issue
#2703

## Motivation and Context
Fixes release issue of ios apps

## How Has This Been Tested?
Uploaded build to ios without manualy adding the push entitlement in xcode

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -78,6 +78,9 @@
     <config-file target="*-Release.plist" parent="aps-environment">
       <string>production</string>
     </config-file>
+     <config-file target="*.entitlements" parent="aps-environment">
+      <string>development</string>
+    </config-file>
     <source-file src="src/ios/AppDelegate+notification.m"/>
     <source-file src="src/ios/PushPlugin.m"/>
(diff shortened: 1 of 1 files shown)


ankon:
I see that in #2703 you suggest to use 'production', while the PR uses 'development'. Is this something that a developer would have to switch around? If so probably documentation is needed for this change.

regevbr:
It should stay developed due to xcode bugs from what I found

schrapel:
@regevbr Do you have any idea why this happens on one of my apps but not the other? 

There must be a way to do this without hardocoding the development/production string value or at least determine where the conflict comes from.

regevbr:
Sorry, no idea... Maybe one of the developers can answer that?

jaux:
@regevbr Can you confirm this issue is due to Xcode, not due to cordova/cordova-ios/push-plugin?

